PR #22, opened 2017-07-18
Persist all of home, remove $HOME change from wrapper.

kujeger:
Persisting all of home instead of doing a workaround to achieve
essentially the same in the wrapper hopefully makes things a bit
clearer and will also work for games doing things like writing to
"/home/$USER/foo" instead of "$HOME/foo".


TingPing:
We might want to migrate existing files.

@alexlarsson Is there a reason you avoided this method?

nanonyme:
Am I understanding this wrong or does this make the entire wrapper script redundant?

kujeger:
It does, but I thought to leave it in case of future needs. If one wants to do some kind of migration stuff, that would for example need to be put in the wrapper.

alexlarsson:
I didn't actually know persist=. worked. If that work it seems like a nice solution to me.

TingPing:
I do believe data migration is a must, nobody should lose data on an update.

kujeger:
I'll write something for migrating

nanonyme:
I wonder if you should just delete the old stuff after migration or keep it and write into some file migration was already done...

kujeger:
I'm thinking something like
```
if [ -d $HOME/.var/app/com.valvesoftware.Steam/home ]; then
  cp -a $HOME/.var/app/com.valvesoftware.Steam/home/* $HOME/
  rm -r $HOME/.var/app/com.valvesoftware.Steam/home
fi
```

that way all files will be copied and kept, and the old stuff cleaned up.

steam-wrapper
```diff
@@ -1,15 +1,11 @@
...
-if [ ! -d "$_STEAM_HOME" ]; then
-    # Create a custom HOME so that games can write wherever they want and it persists
-    mkdir -p $_STEAM_HOME
-    cd $_STEAM_HOME
-    ln -s ../../../../.local .
-    ln -s ../../../../.steam .
-    ln -s ../../../../.var .
+if [ -d $_STEAM_HOME ]; then
+    export GLOBIGNORE=".:.."
+    cp -a $_STEAM_HOME/* $HOME/
```
nanonyme: Since the script is not running with -e in hashbang, if the copy command for some reason fails, it will still wipe the old home. Is that fine?

kujeger: Good point. I'll do some testing with -e .

nanonyme: If you want debuggability on errors, also consider -x which will basically output everything that's being done into terminal

steam-wrapper
```diff
@@ -1,15 +1,11 @@
...
-
-if [ ! -d "$_STEAM_HOME" ]; then
-    # Create a custom HOME so that games can write wherever they want and it persists
-    mkdir -p $_STEAM_HOME
-    cd $_STEAM_HOME
-    ln -s ../../../../.local .
-    ln -s ../../../../.steam .
-    ln -s ../../../../.var .
+if [ -d $_STEAM_HOME ]; then
+    export GLOBIGNORE=".:.."
```
nanonyme: I think this is a bashism, POSIX shell requires you to first set variable, then export the setted variable. Hashbang declares POSIX shell so this might in the future break

nanonyme (approved):
Looks good to me

PR #24, opened 2017-07-24
Build 32-bit version

nanonyme:
64-bit version can't ever work on 32-bit distros. Can also be installed by users with special requirements on 64-bit systems but using 64-bit is highly recommended since it's multilib and can run both 32-bit and 64-bit games

### DIFF
--- a/flathub.json
+++ b/flathub.json
@@ -1,3 +1,3 @@
 {
-  "skip-arches": ["aarch64", "arm", "i386"]
+  "skip-arches": ["aarch64", "arm"]
 }


nanonyme:
Does this support only-arches instead of skip-arches?

TingPing:
You have to make all of this conditional to 64bit: https://github.com/flathub/com.valvesoftware.Steam/blob/master/com.valvesoftware.Steam.json#L22-L34

nanonyme:
Really? I already built this very spec on as i386 on 64bit and it worked fine. I guess need to put this on hold until I'm back at my computer so I can verify. I don't dare to make any more complex changes than this without being able to do local build

TingPing:
It may not cause negative side effects but it is certainly not needed.

nanonyme:
Okay, I'll fix once at home. Any comment on the only-arches? It would be clearer on intent and Flatpak uses it elsewhere but I'm not overly familiar with Flathub details

TingPing:
I don't actually know.

nanonyme:
Hmm, I've been taking a look at the man page remotely. The only way I've thus far been able to find in Flatpak to have different finish-args has been to use separate manifests for 32bit and 64bit as couldn't find conditional statements. I might be missing something here

nanonyme:
As commented on IRC, I don't think that finish-args thing is doable without splitting the setup into a separate multilib app (x86_64 only) and another pure app which has both i386 and x86_64. I'm currently reluctant to do more work on this since it's looking like it'd significantly increase maintenance required

PR #34, opened 2017-10-17
Fdo runtime

valentindavid:
This PR depends on flatpak/freedesktop-sdk-images#51.

Flatpak 0.9.99 is required for using the fd.o runtime. Using flatpak < 0.9.99 should fallback to previous behavior.

Using fd.o runtime should help fix sound problem in various games using alsa-lib. It should make static linking to libstdc++ useless.


TingPing:
I did have to revert 7e999c4 as it would fail to start:

```
Error: You are missing the following 32-bit libraries, and Steam may not run:
libpulsecommon-10.0.so
```

PR #37, opened 2017-10-30
Expand AppData metadata, including screenshots.

mavit:
This additional information should make things look more appealing in, e.g, Gnome Software.

PR #49, opened 2017-12-15
Add OARS metadata. This might be extreme but it's on all Steam.

nedrichards:
I noticed that #37 which added some excellent new appdata info didn't have OARS, so I quickly ran through the [new tool](https://hughsie.github.io/oars/). Steam contains examples of basically everything (as you might expect from the world's most comprehensive PC games store).

Patches and views welcome.

### DIFF
--- a/com.valvesoftware.Steam.appdata.xml
+++ b/com.valvesoftware.Steam.appdata.xml
@@ -31,5 +31,34 @@
       <image type="source" width="1280" height="720">https://raw.githubusercontent.com/flathub/com.valvesoftware.Steam/master/com.valvesoftware.Steam.screenshot-2.png</image>
     </screenshot>
   </screenshots>
+    <content_rating type="oars-1.1">
+    <content_attribute id="violence-cartoon">intense</content_attribute>
+    <content_attribute id="violence-fantasy">intense</content_attribute>
+    <content_attribute id="violence-realistic">intense</content_attribute>
+    <content_attribute id="violence-bloodshed">intense</content_attribute>
+    <content_attribute id="violence-sexual">intense</content_attribute>
+    <content_attribute id="violence-desecration">intense</content_attribute>
+    <content_attribute id="violence-slavery">intense</content_attribute>
+    <content_attribute id="violence-worship">intense</content_attribute>
+    <content_attribute id="drugs-alcohol">moderate</content_attribute>
+    <content_attribute id="drugs-narcotics">moderate</content_attribute>
+    <content_attribute id="drugs-tobacco">moderate</content_attribute>
+    <content_attribute id="sex-nudity">intense</content_attribute>
+    <content_attribute id="sex-themes">intense</content_attribute>
+    <content_attribute id="sex-homosexuality">moderate</content_attribute>
+    <content_attribute id="sex-prostitution">moderate</content_attribute>
+    <content_attribute id="sex-adultery">moderate</content_attribute>
+    <content_attribute id="sex-appearance">intense</content_attribute>
+    <content_attribute id="language-profanity">intense</content_attribute>
+    <content_attribute id="language-humor">intense</content_attribute>
+    <content_attribute id="language-discrimination">intense</content_attribute>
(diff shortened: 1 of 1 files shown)


mavit:
I did wonder whether I should include something like this.  On the one hand you can certainly find such things through Steam.  On the other hand, they're not a part of the Steam client itself.  Would we give this same rating to, say, Gnome Software or Firefox?

nedrichards:
GNOME Software is harder but I think we would to firefox, yes. I mean, that's kind of the upstream discussion at: https://github.com/hughsie/oars/issues/1

TingPing:
It is very strange to rate Steam like that, I think there really needs to be some exception for stores because none of the concepts really apply.

nedrichards:
Yeah, it was partly the thought experiment.

nanonyme:
Is this actually intended to be merged at some point?

TingPing:
I'm still not a fan of shoving OARS on everything since it rarely makes sense on anything but individual *games*. But this probably is the best you can make when you are trying to rate all software ever released on Steam...

nanonyme:
I agree, was just commenting on the fact this has remained last 3 months in uncommented not merged but not declined either state. This can easily remain a year more like that but does that make any sense? :)

TingPing:
> This can easily remain a year more like that but does that make any sense? :)

Sure. *Looks at 5 years worth of his bugs that will never get resolved*

nanonyme:
Steam has builtin parental controls anyway (https://support.steampowered.com/kb_article.php?ref=5149-EOPC-9918) what's the point of OARS if the main purpose of that would be to build parental controls into eg Gnome Software?

jurf:
Hm, though in the light of the new non-curated store policy, I think this is quite valid. It doesn’t matter if it’s not in the client itself, you can easily get access, _through_ the client, to all that. Imagine (unless you have kids ;-) you’re a parent, you wouldn’t want your kid to browse the Steam Store without first setting up the parental control yourself.

nanonyme:
Well, yeah, I think in Android store Steam app is also flagged with "under parental supervision".

nanonyme:
Isn't that also effectively what you meant? It's not really about content as such in this case but whether parent knows that they should be explicitly setting parental policies within the app before letting the child use it

com.valvesoftware.Steam.appdata.xml
```diff
@@ -31,5 +31,34 @@
...
+    <content_attribute id="language-profanity">intense</content_attribute>
+    <content_attribute id="language-humor">intense</content_attribute>
+    <content_attribute id="language-discrimination">intense</content_attribute>
+    <content_attribute id="social-chat">intense</content_attribute>
+    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-audio">intense</content_attribute>
+    <content_attribute id="social-location">none</content_attribute>
+    <content_attribute id="social-contacts">none</content_attribute>
+    <content_attribute id="money-purchasing">intense</content_attribute>
+    <content_attribute id="money-gambling">intense</content_attribute>
```
nanonyme: This looks wrong https://steamcommunity.com/discussions/forum/1/1520386297699215529/

jurf: Hm, are there gambling _games_ on Steam?

nanonyme: No but OARS says gambling includes eg games where you have a spinning wheel that allows you to win in-app credits that can also be bought. It says nothing about needing to turn them back into money which is imo a fundamental part of gambling

nanonyme: Or did you mean something like this https://store.steampowered.com/app/779570/Roulette_Simulator/

Alexander-Wilms: Any game with loot boxes is gambling. At least in belgium: https://arstechnica.com/gaming/2018/04/video-game-loot-boxes-are-now-considered-criminal-gambling-in-belgium

Maybe the OARS definition should be adapted.

PR #68, opened 2018-03-14
com.valvesoftware.Steam.json

LeandroStanger:


nedrichards:
Unless we absolutely know we need this update I'd rather hold back

Issue #70, opened 2018-03-28
Some Feral Interactive games wouldn't run


doraskayo:
It seems that this is actually caused by Feral's launcher, or some version of it. It uses the `ptrace` syscall and exits with code 1 when it's not available, and it's not available by default inside Flatpak's sandbox.

The only game that I own which is published by Feral and uses this launcher is [HITMAN](http://store.steampowered.com/app/236870/HITMAN/), (episode 2 is currently free for a limited time; see the link) so I used it for this analysis. I did find [other reports](https://forum.snapcraft.io/t/solved-unable-to-use-ptrace-in-confinement/3297) which suggest that Feral games rely on this syscall.

It's possible to solve this issue using "--allow=devel", to allow the `ptrace` syscall inside the sandbox (and also others; `personality` and `perf_event_open`. [See here](https://github.com/flatpak/flatpak/blob/3273e9923c49e1effee61d81f8069eef476c5d57/common/flatpak-run.c#L2048).). Going by @alexlarsson's advice, we should probably just enable that for this flatpak. I'll submit a PR soon.

PR #71, opened 2018-03-28
Use "--allow=devel" to allow some blacklisted syscalls inside the sandbox

doraskayo:
This PR enables the `ptrace`, `personality` and `perf_event_open` syscalls inside the sandbox. While we're technically only interested in allowing the `ptrace` syscall for the purpose of solving #70, Flatpak doesn't currently allow granular whitelisting of syscalls, so this is our only option.

See #70 for more information.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -18,6 +18,7 @@
         "--filesystem=xdg-music:ro",
         "--device=all",
         "--allow=multiarch",
+        "--allow=devel",
         "--persist=.",
         "--extension=org.freedesktop.Platform.Compat32=directory=lib/32bit",
         "--extension=org.freedesktop.Platform.Compat32=version=1.6",


nedrichards:
bot, build

nedrichards:
And to clarify - the reason for this is that some games, specifically those made by Feral Interactive, require this syscall to work, right?

And those games are tons of the high profile AAA games available on Linux https://www.feralinteractive.com/en/linux-games/

doraskayo:
@nedrichards, exactly. I'm not entirely sure if it's every game which uses their launcher or a version of the launcher which is used by some of of the games, but I did verify that it at least applies to HITMAN. This is the only game that I own which uses this launcher, so I can't verify this on a broader selection of games. However, from my understanding, all of their newer releases should also use this launcher.

nedrichards:
I'm happy to merge this, given the point is - people want to play games. It'd be nice if someone could drop them an email asking if they could sandbox a little better, but that's fine for now.

cpba:
@nedrichards I did send them an email inquiring about it, they just came back to me with an answer:

> 
> Hi Carles,
> 
> Just to update you on this: we are aware of this and it is intentional. While we intend to find a solution, for the moment flatpak will still need --allow=devel as without it the crash diagnosis and support features are broken.
> 
> Thanks again for getting in touch.
> 
> Best regards,
> Gerard McAuley
> Feral Interactive Support
> http://support.feralinteractive.com

nedrichards:
Thanks for getting in touch, that's a moderately encouraging reply at least.

Issue #89, opened 2018-06-10
Missing bluetooth requirement for Steam Controller


ananace:
With the new Steam Controller over Bluetooth update available, the Steam client requires the - equally new - `--allow=bluetooth` Flatpak option to support using said controller.

nanonyme:
The PR was reverted due to excessive amount of issues. Let's try to re-enable later

ananace:
Understandable, guess that for now people will just have to do with manually overriding it in case they need it.  
`flatpak override com.valvesoftware.Steam --allow=bluetooth` - for reference

nanonyme:
Looks like we can re-enable it, just needs to be well documented that support for 0.10.2 and older is dropped

nanonyme:
Note the root issue is eg Ubuntu isn't shipping 0.10.3 which is a bugfix to their Flatpak. 0.10.4 even exists

nanonyme:
@ananace oops; hit https://github.com/flatpak/flatpak/issues/1935 while trying to make things work nicely

nanonyme:
Ok. There's now significantly large audit trail about this. Let's use local overrides for now and keep this ticket open as a reminder

nanonyme:
I think we're closing the point where we can say Flatpak 1.0.0 recommended, 0.11.3 required and enable this.

nanonyme:
Note #8 probably needs handling to make the experience more friendly

nanonyme:
This will most likely be enabled by default somewhere 2019Q1 when Debian 10 has been released

nanonyme:
Closing as implemented.

PR #90, opened 2018-06-10
Allow bluetooth for Steam Controllers (#89)

ananace:


nanonyme (approved):
Looks plausible to me. Is this a no-op of Flatpak older than 0.11.8?

ananace:
I am honestly not sure, and I have a feeling that the answer might be both 'yes' and 'no' depending of if talking about `flatpak` or `flatpak-builder`.

nanonyme:
I was just discussing this on IRC and looks like old versions of Flatpak do not handle unrecognized allow switches safely https://github.com/flatpak/flatpak/commit/c8c0df3fba458844d99ddd773244467fa4a35408
Support came around 0.10.x

nanonyme:
I'm going to merge this. At worst some people might have to install Flatpak from backports if they're using something ancient like Debian stable

PR #97, opened 2018-06-23
Insert StartupWMClass into desktop entry

ghost:


nanonyme:
Pattern looks odd. Is something missing?

9f565fdc:
The pattern means it will insert a new line containing the string `StartupWMClass=Steam` above the 11th line in the file `steam.desktop`.

TingPing:
Use `desktop-file-edit` rather than `sed`.

9f565fdc:
I didn't know about `desktop-file-edit`, thank you. Updated.

Issue #102, opened 2018-07-01
No audio in several games - can't access alsa-pulse.conf


Pshemas:
So I've thought about bringing the audio issue I'm having with several games into one topic as those seem to have the same root cause.
In short - quite a chunk of games do not have sound and I see this (or very similar) in command line:

```
ALSA lib conf.c:3942:(snd_config_update_r) Cannot access file /etc/alsa-pulse.conf
ALSA lib pcm.c:2501:(snd_pcm_open_noupdate) Unknown PCM default
```
Here's the list of games with this problem that I've tested so far:

**Bastion
Deus Ex: Mankind Divided
Don't Starve
Invisible Inc
MadMax (in the WB start "movie" clip there's sound, but no audio ingame)
Rise of the Tomb Raider
Torchlight 2**

Will keep on testing and will update the list as soon as I find new ones.
Here are my basic system specs:
Kernel: 4.12.14-lp150.12.4-default x86_64 bits: 64
Desktop: Xfce 4.12.4 Distro: openSUSE Leap 15.0
Mobo: ASRock model: EP2C602 serial: 149024860000174
           UEFI: American Megatrends v: P1.80 date: 12/09/2013
CPU(s):    2 Octa core Intel Xeon E5-2670 0s (-HT-MCP-SMP-)
           speed/max: 2600/3300 MHz
Graphics:  Card-1: NVIDIA GM204 [GeForce GTX 980]
           Card-2: NVIDIA GP106 [GeForce GTX 1060 3GB]
           Display Server: x11 (X.Org 1.19.6 ) drivers: nvidia,nvidia
           Resolution: 1920x1080@60.00hz, 2560x1440@59.95hz
           OpenGL: renderer: GeForce GTX 980/PCIe/SSE2
           version: 4.5.0 NVIDIA 396.24
Network:   Card: Intel 82574L Gigabit Network Connection driver: e1000e
Drives:    HDD Total Size: 2360.5GB (43.3% used)
Flatpak 0.10.4

Pshemas:
So here's screenshot showing "no audio" entry in the MadMax settings section.
[](https://i.imgur.com/pkkFmpb.jpg)

Pshemas:
@fenryxo posted a workaround that fixes the problem here:
https://github.com/flatpak/flatpak/issues/1875

TingPing:
New update is out, please confirm it is fixed.

Pshemas:
[...]

PR #106, opened 2018-07-10
Attempt to workaround #102

nanonyme:
Some distros set ALSA_CONFIG_PATH which apparently is leaked into Flatpak. Unset it in the Steam wrapper before running Steam


nanonyme:
Noticed a better way to do this

Issue #60, opened 2018-01-28
Steam Cloud path conflicts breaks game-save syncing


da2x:
TL;DR: Paths are broken because of XDG overwrites. Skip to the code example.

Update: Read [this analysis of this bug](https://www.ctrl.blog/entry/flatpak-steamcloud-xdg) which includes two suggestions for possible fixes. The below ended up serving as drafts for that analysis and is less  less accurate.

---

For most games, Steam Cloud syncs saves and configurations to and from `/home/user/.local/share/studio/gamename`. This is the `LinuxXdgDataHome` data root path in Steam Cloud, which documentation suggests is corresponding to `$XDG_DATA_HOME`. It appears that game developers use `$XDG_DATA_HOME` (default in e.g. Unity3D) but Steam itself doesn’t actually read the `$XDG_DATA_HOME` variable but just hardcodes `$HOME/.local/share/` or `$HOME/.config/`.

---

Many games, like __RimWorld__ , won’t read from `/home/user/.local/share/studio/gamename` where saves are synced by Steam but instead only read from and write to `/home/user/.var/app/com.valvesoftware.Steam/data/studio/gamename`. These won’t be synced and players won’t have access to existing saves.

Some games, like __Cities Skylines__, are synced to `/home/user/.local/share/studio/gamename` and will read from both locations, but only write changes and new saves to `/home/user/.var/app/com.valvesoftware.Steam/data/studio/gamename`. Players have access to existing syncs, but can’t update existing saves or add new ones.

A few games, like __FTL__, are synced to `/home/user/.var/app/com.valvesoftware.Steam/data/studio/gamename` and everything appears to work out of the box. These games seem to be the exception.

---

I tested  with restoring the sandboxed `$HOME` prefix stuff in the wrapper committed by @alexlarsson and reverted by @kujeger. This seems to fix cloud syncing with many games, but broke some games like Cities Skylines.

Combining this special sandboxed `$HOME` variable with having `$XDG_DATA_HOME` point to `/home/user/.var/app/com.valvesoftware.Steam/home/.local/share/` and `$XDG_CONFIG_HOME` point to `/home/user/.var/app/com.valvesoftware.Steam/home/.config/` should fix pretty much every game as Steam and games then should agree on where saves are kept. Should be a workable solution that works for a majority of games (at least Unity3D titles, and I suggest most others too)  — unless anyone has contacts at Valve and can make them fix this upstream?

Here is a suggested patch for `steam-wrapper:

[...]

Issue #66, opened 2018-02-25
Can’t start if symlink in home direcory exists


jurf:
I get this:

    bwrap: Can't make symlink at /home/jurf/Arch: File exists

jurf:
I renamed the thing to `Archlinux`. Now I get this:

```
❯ flatpak run com.valvesoftware.Steam 
bwrap: Can't make symlink at /home/jurf/Archlinux: File exists
```

jurf:
Turns out I can start Steam if I rename it before launching. I can rename it back right after.

jurf:
This is the `info -m` of Steam: https://pastebin.com/6rBP0S9n

As you can see it should only have access to `xdg-music`, but, somehow, it recreates symlinks found in the home directory.

```
~/.var/app/com.valvesoftware.Steam 
❯ cat /var/lib/flatpak/overrides/com.valvesoftware.Steam 
[Context]
filesystems=/data/games/steamapps;
```
```
~/.var/app/com.valvesoftware.Steam 
❯ ls
Archlinux@  cache/  config/  data/  Hudba@

~/.var/app/com.valvesoftware.Steam 
❯ file Archlinux 
Archlinux: broken symbolic link to ../../oldhome/home/juraj/

~/.var/app/com.valvesoftware.Steam 
❯ file ~/Archlinux
/home/jurf/Archlinux: symbolic link to /oldhome/home/juraj/
```

Passing `--nofilesystem=xdg-music` doesn’t do anything.

alexlarsson:
Can you `flatpak run -vv com.valvesoftware.Steam`? It should give us the full bwrap commandline.

jurf:
Here you go, let me know if I trimmed it incorrectly: https://pastebin.com/iFyYcMiY

alexlarsson:
Ok, so, here is what happens:

Due to your filesystem layout the `--filesystem=xdg-music` arguments leads flatpak to expose the symlink `$HOME/Archlinux` (because the music dir is a symlink into this symlink). At this point nothing "normal" has exposed the $HOME path in the sandbox, so flatpak thinks its unintialized and thus part of the root tmpfs, so it creates a symlink there. 

[...]

PR #110, opened 2018-07-24
 Workaround sync issues documented in #60

TingPing:


steam-wrapper
```diff
@@ -8,6 +8,7 @@ if ! [ -s /etc/ld.so.conf ]; then
     export STEAM_RUNTIME_PREFER_HOST_LIBRARIES=0
 fi
 
+# Migration from when $HOME used to be overriden:
```
AsciiWolf: Just small typo: overriden -> overridden :-)

nanonyme:
Considering #66 I'd consider this PR high risk until it's verified symlinks work as expected in this setup

steam-wrapper
```diff
@@ -18,4 +19,38 @@ if [ -d $_STEAM_HOME ]; then
...
+_CONFIG=$_STEAM_BASE/.config
+_DATA=$_STEAM_BASE/.local/share
+_CACHE=$_STEAM_BASE/.cache
+
+mkdir -p $_CACHE $_CONFIG $_DATA
+
+# Migrate any old data:
+if [ ! -L $XDG_CONFIG_HOME ]; then
+  cp -a $XDG_CONFIG_HOME/* $_CONFIG
+  # You can't delete $XDG_CONFIG_HOME/user-dirs.dirs
```
nanonyme: The only thing I'm a bit concerned about with this approach is I have no idea how large user-dirs.dirs is and we're creating an extra copy here due to bind mount. The options basically are
1) Avoid copying bind-mounted things, just created directories where missing (need to immediately restart Flatpak because bind mount happens at startup)
2) Copy over everything. Everything should work from current run too, however, on next run we are bind mounting over directory with the real contents so we're wasting some space

nanonyme:
No go with current migration strategy:
"bwrap: Can't create file at /home/user/.var.app/com.valvesoftware.Steam/config/user-dirs.dirs"

nanonyme:
Another issue noticed: the FOO/* cp fails if there are no files under that directory so the migration script fails with fresh installed Steam

steam-wrapper
TingPing: Its a text file with a few lines, its not a concern.

nanonyme:
Tested
1) Installing old Flatpak app and migrating to new worked fine
2) Incomplete migration where migration was cut at failure moving Steam data directory was resumed on next application start attempt resulting in working application once the data directory moving was properly remedied
3) Installing new Flatpak from scratch worked correctly, migration was done on first start

nanonyme:
I'm going to rebase since master has newer version of Steam

nanonyme:
bot, build

AsciiWolf:
Any news?

nanonyme:
No one has tested the real content merge yet, this is on hold until someone does

da2x:
I don’t have the time to test this properly, but I had a quick look. Games sync up but not down: but I’m not sure why. Some game complains that they can’t sync down too.

unable_to_sync

nanonyme:
@da2x did you use the standalone bundle as "upgrade"? Migration isn't in build yet

da2x:
I just applied this patch to `steam-wrapper` post-install locally. I can see from the directory structure that it's using the new wrapper.

nanonyme:
Crap. That sounds a lot like it can't recover from a sync conflict

nanonyme:
@da2x I've already tested both upload and download and both work. Merge conflict not tested yet

nanonyme:
Also I'm now 100% sure this will not be safe. I just reviewed config from Wasteland 2 and the migration script will simply overwrite any files in .config by files in config

nanonyme:
Also it appears there's a TOC.txt file in Wasteland 2 which is a directory index for save games along with timestamps

nanonyme:
I've now tested this twice in a row by restoring .var/app/com.valvesoftware.Steam from backup with rsync. End result with Wasteland 2 is cloud sync starts working *but* user will not get the choice of selecting between local and cloud data. Local data will just wipe the cloud data. The remaining question is whether this is good enough

nanonyme:
I'm going to try to make a prompt in addition to the workaround

nanonyme:
bot, build

steam_wrapper/steam_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+import shutil
+import errno
+import fnmatch
+import subprocess
+
+STEAM_PATH = "/app/bin/steam"
+STEAM_ROOT = os.path.expandvars("$HOME/.var/app/com.valvesoftware.Steam")
+
+def prompt():
+    subprocess.check_call(["zenity", "--question",
```
TingPing: Could handle this better than just letting it raise?

steam_wrapper/steam_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+        return False
+
+def filter_names(root, names, patterns):
+    _names = []
+    for name in names:
+        if not ignored(os.path.join(root, name), patterns):
+            _names.append(name)
+    return _names
+
+def copytree(source, target, ignore=None):
```
TingPing: I imagine `shutil.copytree()` wasn't the behavior you wanted?

nanonyme: It has a deficiency that if target directory exists, it raises exception and no way to avoid it. There's an open PR about this which might eventually get into a later Python release.

steam_wrapper/steam_wrapper.py
nanonyme: Sure, no problem. What should it do, skip migrations and start Steam?

steam_wrapper/steam_wrapper.py
nanonyme: Specifically https://github.com/python/cpython/pull/8792

steam_wrapper/steam_wrapper.py
nanonyme: Another thing, should we prompt for migration on next startup or record somewhere that we already asked and user said no?

nanonyme:
bot, build

nanonyme:
@AsciiWolf interested in reviewing?

nanonyme:
bot, build

AsciiWolf:
@nanonyme I will see if I get some free time to test this this weekend.

da2x:
2ee970f seems to have done the trick. I’m now seeing sync conflicts after upgrading with migration.

AsciiWolf:
@nanonyme I finally got some time to test this. I already had Steam and a few games installed so I tried running the version from this PR and all the save games were correctly migrated. There were no sync conflict dialogs, however I didn't test any case where files in the cloud were newer than the local ones and any new save games that I created afterwards were correctly synced into the cloud.

steam_wrapper/steam_wrapper.py
```diff
@@ -0,0 +1,143 @@
...
+                           "This is com.valvesoftware.Steam cloud sync repair. "
+                           "If you have conflicting local and cloud data for "
+                           "your game, this may result in partial loss of your "
+                           "cloud data. If you instead prefer ensuring cloud data "
+                           "persists, please relocate your "
+                           "~/.var/app/com.valvesoftware.Steam/data/Steam "
+                           "to a secure location, "
+                           "remove ~/.var/app/com.valvesoftware.Steam "
+                           "and put Steam data directory back to avoid needing to "
+                           "re-download games")])
```
AsciiWolf: Small typo: Missing dot at the end of the sentence. ;-) Also, the window has an empty (default) icon which looks bad. Would it be possible to replace it with a question icon?

PR #129, opened 2018-08-13
Update to Freedesktop SDK 18.08

valentindavid:


nanonyme:
Remove the automatic Python3 conversion, it can generate programs that break in subtle ways

TingPing:
> I thought gyp was added together with newer nss, it might be removable

Indeed.

com.valvesoftware.Steam.json
```diff
@@ -20,9 +20,12 @@
         "--allow=multiarch",
         "--allow=devel",
         "--persist=.",
-        "--extension=org.freedesktop.Platform.Compat32=directory=lib/32bit",
-        "--extension=org.freedesktop.Platform.Compat32=version=1.6",
-        "--extension=org.freedesktop.Platform.GL32=directory=lib/32bit/lib/GL",
+        "--extension=org.freedesktop.Platform.Compat.i386=directory=lib/i386-linux-gnu",
+        "--extension=org.freedesktop.Platform.Compat.i386=version=unstable",
```
nanonyme: Is this intentionally unstable?

valentindavid: That is a mistake. Nice you saw it. I had unstable also installed.

nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
Looks good to me. This should also suppress one source of confusion people have had with the app where they run into problem X and think it has something to do with dpkg

Issue #130, opened 2018-08-14
Persistent crashes with cached shaders with Mesa


vibolvireak:
#### Environment

| Operation System | Flatpak Version | Desktop Environment |
| --- | --- | --- |
| Fedora 28 | 0.99.3 | Gnome 3.28.2

```shell
$ flatpak run com.valvesoftware.Steam -d
```
```
Running Steam on org.freedesktop.platform 18.08 64-bit
STEAM_RUNTIME is enabled automatically
Pins up-to-date!
Installing breakpad exception handler for appid(steam)/version(1533766730)
Installing breakpad exception handler for appid(steam)/version(1533766730)
Installing breakpad exception handler for appid(steam)/version(1533766730)
sh: lspci: command not found
Installing breakpad exception handler for appid(steam)/version(1533766730)
surface_linux.cpp (2095) : Assertion Failed: None of the authentication protocols specified are supported
Assert( Assertion Failed: None of the authentication protocols specified are supported ):surface_linux.cpp:2095

Installing breakpad exception handler for appid(steam)/version(1533766730)
assert_20180814090255_6.dmp[161]: Uploading dump (out-of-process)
/tmp/dumps/assert_20180814090255_6.dmp
[0814/090256.298400:WARNING:crash_reporting.cc(254)] Failed to set crash key: UserID with value: 0
[0814/090256.298490:WARNING:crash_reporting.cc(254)] Failed to set crash key: BuildID with value: 1533764000
[0814/090256.298500:WARNING:crash_reporting.cc(254)] Failed to set crash key: SteamUniverse with value: Public
[0814/090256.298509:WARNING:crash_reporting.cc(254)] Failed to set crash key: Version with value: 20180808213320
[0814/090256.298517:WARNING:crash_reporting.cc(254)] Failed to set crash key: ProductName with value: cefwebhelper
[0814/090256.298530:WARNING:crash_reporting.cc(254)] Failed to set crash key: Vendor with value: Valve
[0814/090256.373070:ERROR:gpu_process_transport_factory.cc(1029)] Lost UI shared context.
Installing breakpad exception handler for appid(steam)/version(1533766730)
Installing breakpad exception handler for appid(steam)/version(1533766730)
Installing breakpad exception handler for appid(steam)/version(1533766730)
CAppInfoCacheReadFromDiskThread took 81 milliseconds to initialize
Installing breakpad exception handler for appid(steam)/version(1533766730)
[...]

PR #131, opened 2018-08-14
Disable Mesa shader cache as crash workaround

nanonyme:
Add workaround for #130 
Apparently at least with some AMD GPU's Mesa repeatedly generates shaders that when loaded from cache result in crashes. Disable shader cache completely until we have a better workaround.


nanonyme:
bot, build

PR #138, opened 2018-08-19
Re-enable Mesa shader path and add a better workaround

nanonyme:
Conserning #130 ; shader cache is now enabled but flushed when Steam is first started. Proper detection logic for whether cache should be flushed to be done later


nanonyme:
bot, build

PR #147, opened 2018-08-20
More elaborate way of flushing Mesa shader cache

nanonyme:
It's not fun to flush cache on every start of Steam. Instead
link flushing to runtime/extension changes
#130 
@valentindavid


nanonyme:
bot, build

nanonyme:
bot, build

PR #151, opened 2018-08-22
Use Freedesktop Sdk for the Runtime

alatiera:
With the launch of Steam Play Proton, Steam has a dependency again
on Python2. Having to change from .Platform to .Sdk is painful,
but here are couple points about doing it:

* I could not find a recipe for bundling python2 quickly and Steam
is one of the most popular Flatpaks

* I am not sure if Photon introduced any other dependencies with
all the DX to Vulkan translation stuff it does

* Bundling python2 can be a security risk and increases maintenance,
so Its probably better to reuse the one from the .Sdk.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -1,6 +1,6 @@
 {
     "app-id": "com.valvesoftware.Steam",
-    "runtime": "org.freedesktop.Platform",
+    "runtime": "org.freedesktop.Sdk",
     "runtime-version": "18.08",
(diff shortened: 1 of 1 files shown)


nanonyme:
I don't know if this even works but this isn't how you use Flatpak normally and installs loads of deps unnecessarily for users. @TingPing @valentind

nanonyme:
Closing, tracking issue #152

TingPing:
Long term Python 2 should just be bundled. The SDK is massive.

PR #157, opened 2018-08-22
Override consent to false again

nanonyme:
Realized that *if* Flatpak was to ever point XDG_CONFIG_HOME to .config, the migration would delete basically everything
Ping @TingPing


TingPing:
> XDG_CONFIG_HOME to .config

Why would it do that? I don't get your point.

nanonyme:
False alarm, the problem was steam-helper relied on cwd without doing chdir.

PR #165, opened 2018-08-24
Add lightweight terminal emulator

gasinvein:
Currently, inspecting a game's console output requires running game's binary directly, with library path pointing to steam runtime, from flatpak app shell, which is way too complicated.
Adding a terminal emulator could help with this a lot, as it makes possible to run game in a terminal window just by setting game's launch options to something like `xterm -e "%command%"` via game settings.
Xterm is obvious choice, but maybe not the best.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -68,6 +68,57 @@
 		"sha256": "c1cfd4e1d4d708c031d60801e527abc9b6d34b85f2ffa2cadd21f75ff38151cd"
 	    }]
 	},
+        {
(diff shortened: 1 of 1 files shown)


nanonyme:
No, it doesn't.
Run terminal, then
flatpak run com.valvesoftware.Steam in terminal

gasinvein:
Does Steam always print game's output to console?

nanonyme:
For what I've seen, yes. I'll also provide steamcmd at some point, no time now

gasinvein:
Then terminal is really not needed.

PR #171, opened 2018-08-25
feat: Include lspci to stop Steam complaining

### LINKED ISSUE #164: sh: lspci: command not found
When starting steam from the command line, you can see that steam tries to use `lspci`, but does not find it:

```
Overriding TZ to Europe/Berlin
PATH: /home/user/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Proton 3.7
Running Steam on org.freedesktop.platform 18.08 64-bit
STEAM_RUNTIME is enabled automatically
Pins up-to-date!
Installing breakpad exception handler for appid(steam)/version(1535062307)
Installing breakpad exception handler for appid(steam)/version(1535062307)
sh: lspci: command not found
Installing breakpad exception handler for appid(steam)/version(1535062307)
[...]
```

And later:
```
[...]
Failed to init SteamVR because it isn't installed
sh: lspci: command not found
ExecCommandLine: "'/home/user/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam'"
System startup time: 17,66 seconds
Generating new string page texture 79: 1024x256, total string texture memory is 1,49 MB
[...]
```

I don't know if this has any implications though.

EspadaV8:
Close #164

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -143,16 +143,16 @@
         },
         {
             "name": "lspci",
-            "buildsystem": "simple",
+            "no-autogen": true,
+            "make-install-args": [
+                "PREFIX=/app",
+                "SBINDIR=$(PREFIX)/bin"
+            ],
             "sources": [{
                 "type": "archive",
                 "url": "https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-3.6.2.tar.gz",
                 "sha256": "d5f9254f27bbda8243b345633e980144e6bd2af9c786bb8a152b904530aef599"
-            }],
-            "build-commands": [
-                "make",
-                "cp lspci /app/bin/lspci"
-            ]
+            }]
         }
     ]
 }
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -140,6 +140,19 @@
                 "cp com.valvesoftware.Steam.appdata.xml /app/share/appdata",
                 "install -Dm644 ld.so.conf /app/etc/ld.so.conf"
             ]
+        },
+        {
+            "name": "lspci",
+            "buildsystem": "simple",
+            "sources": [{
+                "type": "archive",
+                "url": "ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-3.6.1.tar.gz",
+                "sha256": "69a555155c9761ad826e74869b2037d0c4c89f52633822aea16317afd48eaab6"
+            }],
+            "build-commands": [
+                "make",
+                "cp lspci /app/bin/lspci"
+            ]
         }
     ]
 }
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -146,8 +146,8 @@
             "buildsystem": "simple",
             "sources": [{
                 "type": "archive",
-                "url": "ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-3.6.1.tar.gz",
-                "sha256": "69a555155c9761ad826e74869b2037d0c4c89f52633822aea16317afd48eaab6"
+                "url": "https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-3.6.2.tar.gz",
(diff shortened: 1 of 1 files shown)


EspadaV8:
As mentioned in #164 I initially tried to do the following to build `lspci`

```
{
    "name": "lspci",
    "sources": [{
        "type": "archive",
        "url": "https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-3.6.2.tar.gz",
        "sha256": "d5f9254f27bbda8243b345633e980144e6bd2af9c786bb8a152b904530aef599"
    }]
}
```

But that gives the error

```
Building module lspci in /home/espadav8/Workspace/com.valvesoftware.Steam/.flatpak-builder/build/lspci-5
========================================================================
Error: module lspci: Can't find autogen, autogen.sh or bootstrap
```

If there's a nicer way to build this rather than `make`-ing and `cp`-ing the file into the correct place I'd be more than happy to change it (that's part of the reason I didn't open the PR in the first place, because it just seemed 'wrong').

nanonyme:
make PREFIX=/app install
Or use install command to put file in place with right permissions, probably

com.valvesoftware.Steam.json
```diff
@@ -140,6 +140,19 @@
...
+            "name": "lspci",
+            "buildsystem": "simple",
+            "sources": [{
+                "type": "archive",
+                "url": "https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-3.6.2.tar.gz",
+                "sha256": "d5f9254f27bbda8243b345633e980144e6bd2af9c786bb8a152b904530aef599"
+            }],
+            "build-commands": [
+                "make",
+                "cp lspci /app/bin/lspci"
```
EspadaV8: Would this be better done by `install lspci /app/bin/lapci`? Would it make much/any difference?

nanonyme:
Install also takes eg - - chmod=755
Also there's apparently -s for stripping

EspadaV8:
Ah, @nanonyme, that just reminded me. `make install` attempts to copy `lspci` into `/sbin` (or `/app/sbin` with the prefix) which isn't part of the `PATH` so it can't be found there. Which is the better solution here, `install`-ing the file into `/app/bin` or adding `/app/sbin` into the `PATH`?

nanonyme:
There's by the way also "no-autogen": true, that might allow this to work with default build system

nanonyme:
I guess simple is fine, use explicit perms though since this is an executable

EspadaV8:
@nanonyme got it working with the default build system, `no-autogen` and setting a couple of the `make-install-args`. Everything is now installed into the correct place and should have the right permissions

```
bash-4.3$ ls -la /app/bin
total 352
drwxr-xr-x 2 espadav8 espadav8   4096 Jan  1  1970 .
drwxr-xr-x 6 espadav8 espadav8   4096 Aug 25 21:42 ..
-rwxr-xr-x 4 espadav8 espadav8  51272 Jan  1  1970 cmp
-rwxr-xr-x 4 espadav8 espadav8 115608 Jan  1  1970 lspci
-rwxr-xr-x 4 espadav8 espadav8  14712 Jan  1  1970 notify-send
-rwxr-xr-x 4 espadav8 espadav8  52376 Jan  1  1970 setpci
-rwxr-xr-x 4 espadav8 espadav8   5498 Jan  1  1970 steam
-rwxr-xr-x 4 espadav8 espadav8  11384 Jan  1  1970 steamdeps
-rwxr-xr-x 4 espadav8 espadav8    456 Jan  1  1970 steam-wrapper
-rwxr-xr-x 4 espadav8 espadav8   1755 Jan  1  1970 update-pciids
-rwxr-xr-x 4 espadav8 espadav8  16982 Jan  1  1970 xkeystone
-rwxr-xr-x 4 espadav8 espadav8  60144 Jan  1  1970 xrandr
```

nanonyme:
bot, build

nanonyme:
https://flathub.org/builds/#/builders/2/builds/3840

EspadaV8:
Wow, that was done a lot quicker than I expected. Thanks @nanonyme :smile:

Issue #77, opened 2018-04-17
Support for Gamemode


LeandroStanger:
https://github.com/FeralInteractive/gamemode

nanonyme:
I'm not sure this is reasonable to do on Flatpak-level as it doesn't know when Steam has launched a game. This seems more like a thing that should be in Steam itself.

nanonyme:
I guess something that could be done with this is to ship libgamemode in the app and expose gamemode DBus to the container so the daemon gets to run in host and games that grow integration could take it into use. @TingPing opinions?

nanonyme:
Seems a bit strange the library would need systemd headers. I thought they were just for the daemon

nanonyme:
@Tomin1 Try passing it with-systemd=false with-daemon=false

nanonyme:
Please make a pull request, we can work on review and improvements there

valentindavid:
I suppose that to work you need to install game mode daemon on the host. Are there compatibility issues across versions?

valentindavid:
This is something that would be good for non steam games on Flathub. Should it go in the shared modules?

nanonyme:
Imo let's do it right this time and only put it to shared modules once we have proof it works with the build system. It's easier to test build embedded modules

Tomin1:
@nanonyme Thank you for your insight. It was helpful although not enough here. I made a pull request.

LeandroStanger:
@nanonyme Gamemode activated automatically?

nanonyme:
No, you can do the described LD_PRELOAD trick manually if you need it. The library is under /app/lib. Caveat is we cannot currently provide it at all for 32bit games due to build system limitations. Note that the preload trick will not work for all games anyway

nanonyme:
I guess we could add easier variable for setting the preload still so you wouldn't need to know the paths

jurf:
@nanonyme is that the exact launch option you use? I was wondering whether it shouldn’t be `LD_PRELOAD=$LD_PRELOAD:/app/lib/libgamemodeauto.so.0 %command%` for Flatpak. Also, do you have a game where you can confirm it works so I can try too?

nanonyme:
No. I don't think it even does anything on much on my machine as I have AMD CPU

nanonyme:
Might have broken when we removed full system bus access. Metadata only gives permissions to session bus which is probably wrong

nanonyme:
[...]

PR #173, opened 2018-08-25
Add Feral GameMode as module

Tomin1:
Per https://github.com/flathub/com.valvesoftware.Steam/issues/77.

This adds GameMode version 1.2 as a module. **It doesn't build yet** because
it requires libsystemd headers that are not available. GameMode requires libsystemd (it's not optional), particularly systemd/sd-bus.h header. See https://github.com/FeralInteractive/gamemode/issues/15.

You can get this build a bit further by changing line 39 on meson.build to have required: false (with a patch for example) but it will still fail when building. I also think it uses libsystemd to communicate which means that the library needs to be present.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -139,9 +139,9 @@
 	        "-Dwith-examples=false"
 	    ],
 	    "sources": [{
-	        "type": "git",
-	        "url": "https://github.com/FeralInteractive/gamemode.git",
-	        "tag": "1.2"
+	        "type": "archive",
+	        "url": "https://github.com/FeralInteractive/gamemode/releases/download/1.2/gamemode-1.2.tar.xz",
+	        "sha256": "a7b8d63ffdcbea0dc8b557fda42a9471fa9ab0961a5450d2a15cccca0aaf6a95"
 	    }]
 	},
 	{
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -15,6 +15,7 @@
         "--talk-name=org.gnome.SettingsDaemon",
         "--talk-name=org.freedesktop.NetworkManager",
         "--talk-name=org.kde.StatusNotifierWatcher",
+        "--talk-name=com.feralinteractive.GameMode",
         "--filesystem=xdg-music:ro",
         "--filesystem=xdg-pictures:ro",
         "--device=all",
@@ -129,6 +130,20 @@
                 "/share/pkgconfig"
             ]
         },
+	{
+	    "name": "gamemode",
+	    "buildsystem": "meson",
+	    "config-opts": [
+	        "-Dwith-systemd=false",
+	        "-Dwith-daemon=false",
+	        "-Dwith-examples=false"
+	    ],
+	    "sources": [{
+	        "type": "git",
+	        "url": "https://github.com/FeralInteractive/gamemode.git",
+	        "tag": "1.2"
+	    }]
+	},
 	{
 	    "name": "steam_wrapper",
 	    "buildsystem": "simple",


nanonyme:
bot, build

nanonyme:
https://flathub.org/builds/#/builders/2/builds/3844

nanonyme:
@valentindavid opinions?

nanonyme:
Asked some defining questions in linked issue

nanonyme:
I don't think the one header is enough, the build script detects it through pkgconfig.

nanonyme:
Ok. We just discussed this, we probably need to bundle libsystemd in the app as a result of this. *However*, the "do not compile daemon" part still applies

nanonyme:
May need to grab relevant stuff from here https://www.freedesktop.org/software/systemd/

gasinvein:
Why not just build systemd? Without binaries and units stuff, it would be pretty small.

nanonyme:
Yes, that's what I was trying to say. Libsystemd is released as part of systemd, apparently

nanonyme:
It's by the way also possible to use tarballs from https://github.com/FeralInteractive/gamemode/releases for gamemode; the build system uses fairly efficient caching for HTTP downloads
Also you get to verify that what you download doesn't change as you can set sha256 hash for tarball.

nanonyme:
The CI system apparently most likely doesn't preserve Git repos so using Git directly means full clone on every build from Github

nanonyme:
We just had a discussion: to prevent forgeries if you really need direct Git, either hash or both tag and hash need to be used. I think in this case we don't need direct Git and it might even be more efficient for the builder infrastructure.

Tomin1:
@gasinvein I'll check that pull request tomorrow and merge it.

@nanonyme I used git just because "Source code" on that Releases page didn't have inih subproject (submodule) included. Apparently I could have used those .tar.xz packages instead. I will change that.

Should I squash new changes or do you want to preserve history?

gasinvein:
Shoud I reopen PR against this repo?

nanonyme:
Please do, otherwise we can't really know if the builder likes it or not

nanonyme:
Preferably without changes from @Tomin1

gasinvein:
@nanonyme opened #178

nanonyme:
bot, build

nanonyme:
Fun. So looks like merging the PR's resulted in unexpected order in the build flow. See https://github.com/flatpak/flatpak/issues/568

nanonyme:
bot, build

nanonyme:
@Tomin1 please rebase in the morning

nanonyme:
(it would probably work but as I was disappointed to find out, test build does *not* test actual merged result but instead just the PR)

gasinvein:
@nanonyme this PR merges and builds just fine. Why need rebase?

Tomin1:
Rebased on top of flathub:feature/gamemode. I hope this is what you wanted.

This doesn't build on my computer because the systemd build fails for some reason. It's a bit odd.

gasinvein:
>This doesn't build on my computer because the systemd build fails

@Tomin1 that's strange. Can you please provide build log?

Tomin1:
@gasinvein [build-log.log](https://github.com/flathub/com.valvesoftware.Steam/files/2323125/build-log.log)

gasinvein:
I can't find any difference here with my or bot's log, except for lines order. Can you try a single-threaded build (`--jobs 1`)?
Also, I use, flatpak-builder 0.99.3, maybe it matters.

nanonyme:
If it's a concurrency issue, we should probably set that switch in manifest.

gasinvein:
Are you sure? I guess all our build were concurrent, and I also did local build with different number of threads, without errors.

nanonyme:
I did actually hit that same error locally but not through Flathub build system. I assumed the issue was related to me running Flatpak variant of flatpak-builder

gasinvein:
Did you try non-flatpacked flatpak-builder?

nanonyme:
No, was too hard to install at that point to Silverblue due to 1.0.0 release as Flatpak is a protected package there

nanonyme:
The machine is 8 core though so might well provoke concurrency issues

gasinvein:
> The machine is 8 core though so might well provoke concurrency issues

I have 8 logical cores, too. Also tried with 1 and 4 concurrent jobs.

nanonyme:
@Tomin does setting sysconfigdir=/app/etc help any?

gasinvein:
> does setting sysconfigdir=/app/etc help any?

Alredy suspected that and added appropriate two lines. Should I open new PR?

nanonyme:
Nah, no point if you can't reproduce it anyway

Tomin1:
@nanonyme Didn't make any difference. I don't think it's a concurrency issue either, --jobs 1 doesn't help.

nanonyme:
bot, build

nanonyme:
I just realized this might only be half of the solution. We're currently only building 64bit gamemode and libsystemd

Tomin1:
True. Although probably all games that may really benefit from gamemode are 64 bit.

nanonyme:
I'm fine with this and if anyone complains, we can add the 32bit one. I'm not 100% sure if we have a proper environment for cross-compiling per-app 32bit things anyway at this point.

gasinvein:
> I just realized this might only be half of the solution. We're currently only building 64bit gamemode and libsystemd

Interesting point. Is there any good way to bundle foreign arch libs in an app? I guess we will need an extension.

Tomin1:
I don't think that gamemode builds 32 bit libraries on 64 bit systems at all. I you use bootstrap.sh, it will install libgamemode.so to /usr/lib64.

nanonyme:
There's probably some way to cross-compile. Anyway, the SDK currently doesn't support building 32bit app modules for 64bit apps. Currently talking with runtime maintainers on whether that could be made possible since we definitely do have use-cases for it.

NicTanghe:
thesre is definatly no gamemode installed on rhel with flathub

Issue #62, opened 2018-02-07
Portal 2 segfaults when loading any level


Exalm:
```
qhull warning: missing space after flag . (2e); reserved for menu. Skipped.
qhull error: the joggle for 'QJn', 1e-12, is below roundoff for distance computations, 1
qhull warning: unknown flag . (2e)
qhull warning: missing space after flag . (2e); reserved for menu. Skipped.
qhull error: the joggle for 'QJn', 2,4e-12, is below roundoff for distance computations, 1
qhull warning: unknown flag . (2e)

-----snip-------

qhull error: the joggle for 'QJn', 0,00034, is below roundoff for distance computations, 1
qhull warning: unknown flag . (2e)
qhull warning: missing space after flag . (2e); reserved for menu. Skipped.
qhull error: the joggle for 'QJn', 0,0004, is below roundoff for distance computations, 1
qhull warning: unknown flag . (2e)
qhull warning: missing space after flag . (2e); reserved for menu. Skipped.
qhull error: the joggle for 'QJn', 0,00048, is below roundoff for distance computations,
Thread 1 "portal2_linux" received signal SIGSEGV, Segmentation fault.
0xf46c9243 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/bin/vphysics.so
Argument list to give program being debugged when it is started is "-game portal2 -steam".
(gdb) bt
0xf46c9243 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/bin/vphysics.so
Argument list to give program being debugged when it is started is "-game portal2 -steam".
(gdb) bt
#0  0xf46c9243 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/bin/vphysics.so
#1  0xf46c93f8 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/bin/vphysics.so
#2  0xe9ad7fa2 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/portal2/bin/server.so
#3  0xe9ada057 in ?? () from /home/exalm/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Portal 2/portal2/bin/server.so
[...]

PR #176, opened 2018-08-25
Apply workaround for formatting errors

nanonyme:
Investigations in #62 revealed differences in LC_NUMERIC may result in crashes. Force it to be C (generic). Downside is numbers may look non-native but that's not a very high price to pay. 
Can be disabled through flatpak override --env=LC_NUMERIC= com.valvesoftware.Steam


nanonyme:
bot, build

nanonyme:
https://flathub.org/builds/#/builders/2/builds/3853

nanonyme:
@valentindavid discussed this with you earlier

PR #178, opened 2018-08-26
Add libsystemd, which is needed for Feral GameMode

gasinvein:
Build whole systemd and cleanup everything, except for libsystemd and libudev.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -71,6 +71,63 @@
 		"sha256": "c1cfd4e1d4d708c031d60801e527abc9b6d34b85f2ffa2cadd21f75ff38151cd"
 	    }]
 	},
+        {
+            "name": "systemd",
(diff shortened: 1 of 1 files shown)


nanonyme:
bot, build

com.valvesoftware.Steam.json
```diff
@@ -71,6 +71,63 @@
...
+            "cleanup": [
+                "/bin",
+                "/lib/systemd",
+                "/lib/sysctl.d",
+                "/lib/udev/rules.d",
+                "/lib/modprobe.d",
+                "/lib/kernel",
+                "/share/polkit-1",
+                "/include",
+                "/lib/pkgconfig",
```
nanonyme: pkgconfig is used for detection on whether libsystemd is installed so this will break gamemode

nanonyme:
https://flathub.org/builds/#/builders/26/builds/465

PR #185, opened 2018-08-28
Fix mediakeys Dbus name

### LINKED ISSUE #184: Remove access to org.gnome.SettingsDaemon
The correct D-Bus name is `org.gnome.SettingsDaemon.MediaKeys`, and host implementations should be fixed instead of allowing access to a potentially dangerous daemon. See:
https://bugzilla.gnome.org/show_bug.cgi?id=781326#c8

nanonyme:
Closes #184


nanonyme:
bot, build

PR #186, opened 2018-08-28
Temporarily add both keys

nanonyme:
Related to #184 
No time to debug thoroughly which one is actually used by client libraries


nanonyme:
bot, build

PR #187, opened 2018-08-28
use buildsystem: simple

eszlari:


eszlari:
This is a really standard approach. Keep in mind that `buildsystem: simple` was not available back when this manifest was written originally!

nanonyme:
Well, yes but surely simple should not be used for everything. Steam tarball does have a makefile and clear install. However, we're handling multiple completely unrelated files in the same module. Maybe we shouldn't?

com.valvesoftware.Steam.json
```diff
@@ -42,11 +43,6 @@
         "--env=ALSA_CONFIG_PATH=",
         "--env=MESA_GLSL_CACHE_DIR="
     ],
-    "build-options" : {
```
nanonyme: Why was this removed

com.valvesoftware.Steam.json
```diff
@@ -158,7 +154,19 @@
...
+            "build-commands": [
+                "sed -i s:Exec=steam:Exec=/app/bin/steam-wrapper: steam.desktop",
+                "sed -i s:/usr/bin/steam:/app/bin/steam-wrapper: steam.desktop",
+                "desktop-file-edit --set-key=StartupWMClass --set-value=Steam steam.desktop",
+                "sed -i s:/usr/lib/:/app/lib/: steam",
+                "PREFIX=/app make install",
+                "ln -sf /bin/true /app/bin/steamdeps",
+                "cp /usr/bin/cmp /app/bin",
+                "mkdir -p /app/lib/i386-linux-gnu /app/lib/debug/lib/i386-linux-gnu",
+                "install -Dm644 -t /app/share/metainfo com.valvesoftware.Steam.appdata.xml",
```
nanonyme: This doesn't look like simple relocation from default to simple. Could you please split the migration and any functional changes into separate commits?

nanonyme:
bot, build

com.valvesoftware.Steam.json
eszlari: Because it's only needed in the steam module: `"PREFIX=/app make install"`

com.valvesoftware.Steam.json
eszlari: This will not break anything, `/app/share/appdata` is the legacy path, but both are supported, see:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

nanonyme: We're currently supporting quite old versions of Flatpak. Does this break backwards-compatibility and from which versions? Comments in code or commits about this kind of things, please.

eszlari: https://github.com/hughsie/appstream-glib/commit/74f0b2b54890662783ebef64c7f130fa2806a8c0

nanonyme: Sounds too recent for us then

nanonyme: If you want to port to simple, fine but this PR is not acceptable unless you split the commits. It must be possible to say based on commit history why functional changes are done.

eszlari: How is this "recent"? appstream-glib 0.6.7 was released 12 Jan 2017, the 18.08 SDK contains 0.7.10 and there are dozens of apps on Flathub already, that is this path.

nanonyme: You arguing against review comments isn't really productive in getting this merged

nanonyme: Please split the commit so that it can be better followed what gets replaced by what

nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
bot, build

Issue #188, opened 2018-08-28
Get rid of system bus privilege


nanonyme:
Currently Steam has fairly wide access over Dbus. We should drop system bus and make sure that sufficient minimal access is provided. Apparently at least SDL1 might be one of the reasons that cause SDL1 being necessary.

nanonyme:
Logs can be obtained through flatpak run --log-session-bus com.valvesoftware.Steam

nanonyme:
May make sense to just remove it and fix any bugs as they occur. Nothing obvious breaks when removing this

PR #189, opened 2018-08-28
Remove unrestricted system bus access

nanonyme:
As noted in #188, just removing this. Regressions will be handled individually (more explicit accesses may be required)


nanonyme:
bot, build

nanonyme:
bot, build

PR #190, opened 2018-08-29
Fix NetworkManager talk-name

gasinvein:
NetworkManager is on system bus, so there should be `--system-talk-name` instead of `--talk-name`.


gasinvein:
>What was the impact of this bug?

With `--socket=system-bus` nothing. Without it - no actual access to NetworkManager.

nanonyme:
I mean, what broke?

nanonyme:
This is just for historic purposes, I'm going to merge it anyway

gasinvein:
As far as I know, Steam in `-steamos` mode use NM for network settings. In normal mode, it probably just gets info about current connections.

PR #192, opened 2018-08-29
Use add-extensions instead of finish-args

gasinvein:
Solves #183, needs testing.

PR #197, opened 2018-08-29
Revert "Use add-extensions instead of finish-args"

nanonyme:
Reverts flathub/com.valvesoftware.Steam#192
The finish-args are not identical despite JSON including everything
@gasinvein

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -23,6 +23,21 @@
         "--allow=multiarch",
         "--allow=devel",
         "--persist=.",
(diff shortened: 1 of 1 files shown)


nanonyme:
See https://github.com/flatpak/flatpak/issues/1746 about autodelete

gasinvein:
I guess `autodelete` it's `false` by default, as for `org.freedesktop.Platform.Compat.i386` it was unset in finish-args, and it didn't cause problems so far.

nanonyme:
I'll confirm first. I was too trigger-happy with the PR as it looked fine, should have looked that up first

PR #200, opened 2018-08-30
Add SteamCMD

gasinvein:
I'm not sure if SteamCMD should be provided as extension, but as most users don't need it, it's better to make it optional.
Todo:
- [x] Reuse steam-wrapper.py with steamcmd
- [x] Make it so that steamcmd desktop entry appears only if extension is installed
- [x] Add appdata for SteamCMD
- [x] Add different icon for SteamCMD


nanonyme:
Something to consider with steam-wrapper:
1) Throw the setup.py stuff away, use regular executable
2) Make it instead of always pointing to steam have first argument be the command
3) Make entrypoint be a shell-script that will call steam-wrapper with steam as first argument

gasinvein:
I thought of importing some functions from steam-wrapper directly. Is it bad idea?

nanonyme:
I dunno. My thought experiment was basically making it use the first argument it gets as parameter. This would allow using it like 
#!/app/bin/steam-helper bash
exec steam $*

nanonyme:
That would basically make it so that migration would be run before following script and all environment variables would be set

nanonyme:
Maybe. I've just been leaning towards @TingPing's opinion that maybe entangling setuptools and pip into all of this may have been overkill. The main reason why there's a special copytree and a lot of other things (instead of some bash as originally) is the Cloud Sync fix was quite complex to get right and required specifically set timestamps in specific places. I dunno. I think it feels strange to have main functions calling each other. I think at this point it might be cleaner if the bootstrap and execution logic was split into a separate helper that would always take a path and the entrypoint of steam-wrapper would call that with path to steam

nanonyme:
(plus yeah, I expected Cloud Sync would go wrong and there might need to be remediation migrations to fix things up which would have been painful in bash)

gasinvein:
I agree with this, but it's feels out of scope here. Maybe we should create another branch with steam_wrapper.py rework?

nanonyme:
Why would steamcmd have a desktop icon, btw? I thought it was a command line tool

gasinvein:
Easy way to launch it. It has it's own command-line prompt, so I find useful ability to launch it in a terminal window from desktop menu.

nanonyme:
Unless I'm misunderstanding something badly, isn't what you created a module, not extension? (https://blog.tingping.se/2018/03/18/flatpaking-plugins.html)

gasinvein:
It's both. SteamCMD is built as a module, then exported separately from main app as an extension.

gasinvein:
@nanonyme Did you mean someting like [this](https://github.com/flathub/com.valvesoftware.Steam/compare/master...gasinvein:wrapper-rework) should be done with steam-wrapper?

nanonyme:
Yeah, something like that. It was originally just another file in steam module, it could go back there. Except instead of separate steam_wrapper.sh could as well use script source

nanonyme:
Anyway. Strange. Looks like when autodownload is disabled, the module is fully installed except without the source tarball downloading. Hence the desktop icon is installed too

gasinvein:
>instead of separate steam_wrapper.sh could as well use script source

Can yo explain it, please? My english not very good.

nanonyme:
{
  "type": "script",
  "dest-filename": "steam-wrapper.sh",
   "commands": ["#!/app/bin/steam-helper bash", "exec $*"]
}
Or somesuch

nanonyme:
Afaik the syntax is designed for doing trivial snippets without needing separate files

gasinvein:
Is it possible? I thought flatpak-builder always appends `#!/bin/sh` to script type sources.

nanonyme:
Started doubting that as you said. Anyway, I'm still not quite sure you can do that conditional install of desktop file

steam_wrapper/steamcmd_wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+    steamcmd_path = os.path.join(
+        steamcmd_install_dir, "steamcmd.sh"
+    )
+
+    if not os.path.isfile(steamcmd_path):
+        if os.path.isfile(os.path.join(STEAMCMD_BOOTSTRAP_DIR, "steamcmd.sh")):
+            steam_wrapper.copytree(
+                STEAMCMD_BOOTSTRAP_DIR,
+                steamcmd_install_dir,
```
nanonyme: Doesn't this effectively mean uninstalling the app will not remove steamcmd?

gasinvein: Yes. But I don't see any way to avoid this.

nanonyme: I see what you mean.

nanonyme:
If you can't get the conditional download and conditional desktop to work, I suggest you either download always or remove the entire desktop file. The logic you want would most likely easily work with a real app extension

nanonyme:
I think we can as well just drop the extension thing if it's providing too much trouble. It's not as if steamcmd is that massive anyway. It's like just 6MB

gasinvein:
What if we symlink `/app/share/applications/steamcmd.desktop` to `/app/steamcmd/share/applications/steamcmd.desktop`? Thus without extension installed the desktop file will be a broken symlink. Will it break anything, e.g. desktop menu?

nanonyme:
I don't think /app/steamcmd/share/applications/steamcmd.desktop has anything to do with it, it's the desktop file that you install in build-commands that creates the visible desktop entry

gasinvein:
I missed that, what's wrong with conditional download?

nanonyme:
Well, with the conditional download get quickly exiting prompt with failure if the cmd is not installed if you try to open the start menu entry. It would imo be more consistent that either the start menu entry does something useful or doesn't exist at all

nanonyme:
I'd say extension. Let's just keep it simple

gasinvein:
Done. Now we need new appdata I guess.

nanonyme:
Is there some generic free to use terminal icon that could be used? Anything Steam-related is most likely heavily trademarked so we should only use pristine images

gasinvein:
Does flatpak require icon to be packaged in bundle? If not, we can use e.g. generic `terminal` icon or so.

nanonyme:
I would assume it is not allowed as it imposes a dependency to host

gasinvein:
I doubt there is a single linux desktop without a terminal icon.

gasinvein:
`utilities-terminal` icon is present in all default icon themes, including HighContrast. I thing we can rely on it.

nanonyme:
Sounds fine to me

nanonyme:
bot, build

AsciiWolf:
Sorry, but I really dislike that SteamCMD is installed by default as a part of the Steam Flatpak. It definitely should be an addon or a separate Flatpak.

nanonyme:
Addons don't really work apparently, base apps and inheritance might do it

gasinvein:
@AsciiWolf do you dislike desktop entry or the very presence of steamcmd?

AsciiWolf:
@gasinvein The desktop entry.

nanonyme:
I'm fine either way for desktop entry. You can make a PR if you like. Steamcmd itself has been requested for inclusion before this. Note that I cannot currently test any PR's due to broken machine

AsciiWolf:
[Here](209) is a PR to remove the desktop file.

PR #201, opened 2018-08-30
Revert "Game mode system bus"

nanonyme:
Reverts flathub/com.valvesoftware.Steam#191
@jurf commented in #77 that the original bus was right


nanonyme:
bot, build

nanonyme:
I'm keeping this on hold since it's not clear whether we can support gamemode

PR #204, opened 2018-09-01
Add bug report template

nanonyme:


nanonyme:
Automating labeling is apparently currently not possible, see https://github.community/t5/How-to-use-Git-and-GitHub/Apply-labels-automatically-at-issue-creation/m-p/10749#M3509

nanonyme:
Also ping @valentindavid on what information we should ask by default to make bugs easier to debug.

.github/ISSUE_TEMPLATE/game_issue.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+**Game information**
+<!-- Name of the game or games that do not work -->
+
+**Distribution name**
+<!-- If your distribution is versioned, incldue the version. Also preferably last update time -->
+
+**Flatpak info**
+<!-- Flatpak version -->
```
EspadaV8: Perhaps include a command to run that includes the information you're interested in. Is just `flatpak --version` enough? Perhaps `flatpak --show-details --all` too?

.github/ISSUE_TEMPLATE/game_issue.md
```diff
@@ -0,0 +1,18 @@
...
+name: Bug report about game issues
+about: This report type is speficically meant or issues with games
+
+---
+
+**Game information**
+<!-- Name of the game or games that do not work -->
+
+**Distribution name**
+<!-- If your distribution is versioned, incldue the version. Also preferably last update time -->
```
EspadaV8: Perhaps include a command to get the information? `lsb_release --all`?

.github/ISSUE_TEMPLATE/game_issue.md
EspadaV8: Doing `--all` might show a bit too much information since it would contain all the other flatpaks that are installed, is there a way to get just the Steam information and its deps?

nanonyme: I guess could give a simple grep spell that would only give this app's namespace and everything under org.freedesktop.Platform

.github/ISSUE_TEMPLATE/game_issue.md
nanonyme: I have no idea if there's a single way that works everywhere

.github/ISSUE_TEMPLATE/game_issue.md
EspadaV8: `flatpak info com.valvesoftware.Steam` shows a lot of information about the Steam pak but I can't see anything to list the deps of it too.

EspadaV8: `glxinfo | grep OpenGL`? Or does something need to be run from the inside?

nanonyme: Yes, definitely. We want to know which GL drivers are actually in use from runtime extensions

nanonyme: To be honest, we could probably consider bundling glxinfo and vulkaninfo for helping get better information from inside the app

EspadaV8: If this is the case it would probably be worth adding a shell script too and a line here saying something like

Run `flatpak run --command=get-driver-inf com.valvesoftware.Steam` and paste the results here

EspadaV8: My main issue with this template is that it'll be very easy for people to read it and not know exactly what info is needed and will end up either not putting enough/correct information or way too much. If there's anything that can be added to the template and the flatpak then it'd help a lot.

nanonyme: Yeah, this was not intended as a final thing, more as asking for suggestions. My Linux currently currently can't start gnome-shell at all so I've difficulty experimenting which commands to use.

TingPing: `flatpak --gl-driver` is probably good enough for driver version.

.github/ISSUE_TEMPLATE/game_issue.md
TingPing: It is Linux, of course there is no way that works everywhere!

`cat /etc/redhat-release || cat /etc/suse-release || cat /etc/debian-release || cat /etc/lsb_release || cat /etc/os-release`

Who knows.

PR #209, opened 2018-09-02
Remove SteamCMD desktop file

AsciiWolf:


com.valvesoftware.Steam.json
```diff
@@ -215,10 +214,6 @@
                 {
                     "type": "file",
                     "path": "com.valvesoftware.Steam.cmd.metainfo.xml"
```
nanonyme: Do we still need this?

AsciiWolf: Probably no, but it is not causing any harm.

PR #211, opened 2018-09-05
Bugfix/legacy support

### LINKED ISSUE #203: CentOS 7 - Steam Error - glxChooseVisual - failed
Everything was working fine until I updated com.valvesoftware.Steam. Please help.

`glXChooseVisual failed
 glXChooseVisual failedMain.cpp (326) : Assertion Failed: Fatal Error: glXChooseVisual failed
`

nanonyme:
Looks like after migration to Freedesktop runtime 18.08 the legacy path leads into errors like in #203; remove STEAM_RUNTIME_PREFER_HOST_LIBRARIES seems to make the issues go away. This should fix #203

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -136,7 +136,6 @@ def legacy_support():
     if force_legacy or not check_nonempty("/etc/ld.so.conf"):
         # Fallback for flatpak < 0.9.99
         os.environ["LD_LIBRARY_PATH"] = "/app/lib:/app/lib/i386-linux-gnu"
-        os.environ["STEAM_RUNTIME_PREFER_HOST_LIBRARIES"] = "0"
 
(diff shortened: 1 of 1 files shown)


nanonyme:
bot, build

nanonyme:
Will still be testing with an actual old Flatpak before merging

eszlari:
Have you considered using `--require-version` in `finish-args`?

nanonyme:
I don't recommend anyone to use require-version if you're not ready to accept the fact that it may cause flatpak update to crash for your end-users.

nanonyme:
Okay, this is taking more time than expected as I ended up building a VM to easier build arbitrary versions of Flatpak

nanonyme:
Considering 0.8.x support seems to also be broken in other interesting ways, support will probably be removed in one way or several #212

PR #213, opened 2018-09-06
Feature/bluetooth

nanonyme:
Require Flatpak 0.10.3
Verified distributions:
Debian 9 (mitigated through backports, will be fine after Debian 10)
Solus (ok)
Mageia (ok)
CentOS 7 (amigadave COPR can be used for now)

All in all there will be affected users but it's most likely simpler to help upgrading Flatpak than to debug esoteric issues with pre-0.9.99.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -28,7 +28,7 @@
         "--env=LC_NUMERIC=C",
         "--env=ALSA_CONFIG_PATH=",
         "--env=MESA_GLSL_CACHE_DIR=",
-	"--require-version=0.10.3"
+        "--require-version=0.10.3"
     ],
     "add-extensions": {
         "org.freedesktop.Platform.Compat.i386": {
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -27,8 +27,8 @@
         "--persist=.",
         "--env=LC_NUMERIC=C",
         "--env=ALSA_CONFIG_PATH=",
-        "--env=MESA_GLSL_CACHE_DIR="
-	"--require-version=0.10.3",
+        "--env=MESA_GLSL_CACHE_DIR=",
+	"--require-version=0.10.3"
     ],
     "add-extensions": {
         "org.freedesktop.Platform.Compat.i386": {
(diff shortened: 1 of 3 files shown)


nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
bot, build

steam_wrapper/steam_wrapper.py
```diff
@@ -132,11 +132,10 @@ def check_nonempty(name):
...
 
 def legacy_support():
-    if not check_nonempty("/etc/ld.so.conf"):
-        # Fallback for flatpak < 0.9.99
-        os.environ["LD_LIBRARY_PATH"] = "/app/lib:/app/lib/i386-linux-gnu"
-        os.environ["STEAM_RUNTIME_PREFER_HOST_LIBRARIES"] = "0"
-
+    current_info = read_flatpak_info(FLATPAK_INFO)
+    version = tuple(int(x) for x in current_info["flatpak-version"].split("."))
+    if version < (0, 10, 3):
```
gasinvein: Just use `distutils.version.StrictVersion`, like so:
```python
if StrictVersion(current_info["flatpak-version"]) < StrictVersion("0.10.33"):
    ...
```

com.valvesoftware.Steam.json
```diff
@@ -27,8 +27,8 @@
         "--persist=.",
         "--env=LC_NUMERIC=C",
         "--env=ALSA_CONFIG_PATH=",
-        "--env=MESA_GLSL_CACHE_DIR="
-	"--require-version=0.10.3",
+        "--env=MESA_GLSL_CACHE_DIR=",
+	"--require-version=0.10.3"
```
gasinvein: Here is a tab yet again :)

nanonyme: Oh blergh. Did rebase to master.

nanonyme:
bot, build

nanonyme:
Tested that there in fact was no crash on update with 0.8.6. It might be a later regression.

nanonyme:
I'm going to decline this again and instead work in another branch the groundwork on doing both runtime and requires check that requires 0.9.0. This will then be made on top of that work and require a higher version.

Issue #214, opened 2018-09-08
Shield Steam from timezone mess


nanonyme:
TLDR: Steam timezone handling relies on assumptions about zoneinfo files instead of just UTC deltas being used.

Copy from @tomgar
> The workaround would be better if it interprets a TZ env which references a file. In my case I have set >TZ as :/etc/localtime for good reasons:
>https://blog.packagecloud.io/eng/2017/02/21/set-environment-variable-save-thousands-of-system-calls/
>, being this way of setting up the TZ env part of the spec:
>https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html (search for third format)
>
>In your steam-wrapper, you are almost there... you need to check if the TZ env starts with a colon and >then check if it's a symbolic link and follow it. Either after that or if it's not a symbolic link, just extract >the filename as you are already doing in that for loop.

>Steam doesn't work with this kind of TZ representation so that's why I'm commenting here:
>ValveSoftware/steam-for-linux#5609

>Hopefully Steam will fix this issue in their side some day...

See #120
Essentially we should in all cases normalize input into something that Steam likes whether or not TZ is given in environment.

nanonyme:
So there's a couple cases for TZ

1. Difference from UTC. By far easiest for us to use, I think /etc/localtime has all the relevant information for this
2. Format with daylight saving time. Seems really complex, I would really hope this isn't needed
3. :characters name of the file that contains zoneinfo. This is a bit iffy. That file might not exist in the sandbox
4. :characters that does not begin with slash and refers to zoneinfo file

Seems glibc violates its own documentation though and fourth case is allowed to be present without leading :

nanonyme:
So currently it works so that your host TZ is simply ignored. If you really know what you're doing

flatpak --user override --env=TZ=zoneinfo com.valvesoftware.Steam

works to set different zoneinfo than what the algorithm gives

nanonyme:
@tomgar do you think that's sufficient or do we need real TZ handling?

tomgar:
I was aware I could do that... but that means I'll have to run Steam from command line every time or alter the .desktop file. That doesn't look good for new users. This TZ issue should be transparent to users and prevent users from looking this github for a solution, in my opinion.

It's a pity that the flatpak runtime doesn't come with systemd... you could run: timedatectl show -p Timezone --value and use the output as the TZ env value.

tomgar:
By the way, I appreciate the efforts you're taking with this.

[...]

PR #215, opened 2018-09-08
Mask TZ

nanonyme:
See #214, supporting TZ fully might be reasonably complex given the semantics of it so for now just don't let the host TZ leak into application context without an explicit override. This should make the TZ workaround be used for everyone unless there's an override.


nanonyme:
bot, build

PR #219, opened 2018-09-10
Micro-optimize local builds

nanonyme:
lspci is currently recompiled on every build. Re-order to try to make it be cached more often


nanonyme:
bot, build

PR #230, opened 2018-09-15
Use LooseVersion instead of StrictVersion

### LINKED ISSUE #229: legacy version check broken with flatpak 0.11.8.3
the minimum flatpak version check introduced in https://github.com/flathub/com.valvesoftware.Steam/commit/55f7e814baffcf9c6f0a5f3c33fcc42b35c77730 breaks with e.g. flatpak 0.11.8.3 because `distutils.StrictVersion` only supports versions with 2 or 3 components. See [PEP-386](https://www.python.org/dev/peps/pep-0386/#distutils) for documentation and alternatives. I haven't looked into what the best one is (maybe `LooseVersion` instead or the `setuptools` implementation?) and this PEP is also superseded by [PEP-440](https://www.python.org/dev/peps/pep-0440/) which does not offer an API afaict from a quick glance

nanonyme:
Fixes #229

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -8,7 +8,7 @@ import fnmatch
 import subprocess
 import glob
 import configparser
-from distutils.version import StrictVersion
+from distutils.version import LooseVersion
 
 
 STEAM_PATH = "/app/bin/steam"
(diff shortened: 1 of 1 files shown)


nanonyme:
bot, build

PR #258, opened 2018-11-06
Feature/sdl wayland

### LINKED ISSUE #257: Expose Wayland by default
Currently Wayland socket is not exposed at all. Let's expose it *but* also clear sdl driver variable. This should make Wayland be used with SDL2 when it either becomes default or end user sets override.

nanonyme:
Closes #257 
@TingPing any opinions?

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -31,6 +31,7 @@
         "--env=MESA_GLSL_CACHE_DIR=",
         "--env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES=",
         "--env=STEAM_RUNTIME=",
+        "--env=SDL_VIDEODRIVER=",
         "--require-version=0.9.0"
     ],
     "add-extensions": {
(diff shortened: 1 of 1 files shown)


nanonyme:
bot, build

nanonyme:
bot, build

TingPing:
I have my doubts that there won't be broken games because of this, but I don't know.

nanonyme:
Yeah, that's possible, I guess. Maybe we should have a mechanism for masking the WAYLAND_DISPLAY for badly behaving apps (although basically all breakage I've seen thus far has been software where end-user forces Wayland on even when software doesn't properly support it) before turning this on. Then again, I guess really bad software could also be hardcoding wayland-0 but not really supporting Wayland correctly...

nanonyme:
My idea was the main source of breakage would be end user setting SDL_DISPLAYDRIVER=wayland in their host and it leaking into the app which would result Wayland getting enabled when not really supported.

nanonyme:
I'll leave this open in current form for now, should be easy to rebase if needed.

TingPing:
> Hm, I guess there's also the thing that if some games break with Wayland exposed, they'd most likely currently be broken with native Steam already.

That is true and makes this more reasonable to try out.

nanonyme:
bot, build

PR #259, opened 2018-11-06
Remove gamemode from application

nanonyme:
Since gamemode never really properly worked anyway and Meson broke
systemd builds for all released versions of systemd, just remove
the existing code
See #77


nanonyme:
bot, build

PR #261, opened 2018-11-07
Set default that frame is minimized

### LINKED ISSUE #170: steam "minimize to tray". A bad thing today? [Discussion]
On Linux we have that "problem" of many desktops. Some of them without tray support at all. Like Ubuntus new Main desktop GNOME (but they shipp tray support for now, I think) But this keeps me thinking If we should set the option

`STEAM_FRAME_FORCE_CLOSE=${STEAM_FRAME_FORCE_CLOSE:-0}`
So steam will minimize on close, instead of minimize to tray on close. We already do this for a long time on Manjaros Steam client without users going crazy. So that's not a problem here.

What do you think?

nanonyme:
Closes #170


nanonyme:
bot, build

PR #267, opened 2018-11-30
Refuse to start if flatpak-info contains bad overrides

### LINKED ISSUE #266: Sanity checks into steam-wrapper for filesystem overrides
To fix up cloud sync, steam-wrapper is rewriting user home. This will work fine as long as user home is the sandbox. If someone users --filesystem=home or --filesystem=host or any other way that will expose the real user home as writable, this application may end up breaking that user's home directory. 
As mitigation, steam-wrapper should extract override data from /.flatpak-info and refuse to do anything if dangerous overrides are set.

nanonyme:
Fixes #266


nanonyme:
bot, build

PR #271, opened 2018-12-05
Use SDK CFLAGS for lspci

nanonyme:


nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
bot, build

nanonyme:
bot, build

com.valvesoftware.Steam.json
```diff
@@ -96,6 +96,9 @@
         {
             "name": "lspci",
             "no-autogen": true,
+	    "make-args": [
```
gasinvein: Tab/spaces mix here.
```suggestion
            "make-args": [
```

gasinvein (changes requested):

com.valvesoftware.Steam.json
nanonyme: Sure, that's not the reason why this doesn't work though.

nanonyme:
bot, build

com.valvesoftware.Steam.json
gasinvein: Just noticed that and commented, sorry.

nanonyme: Yeah, it's a good catch. I'm using Emacs and it wants to always put tabs into JSON. Anyhow, completely rewritten, I couldn't manage to get CFLAGS to go through without buildsystem simple (lspci isn't really an autotools project anyway, it just has a Makefile)

Issue #95, opened 2018-06-11
Torchligt 2 - crash on start


Pshemas:
With this one I'm not sure it is flatpak package related as it seems it is not working in general (need to test on a machine with  a configuration I know worked), but just in case here's the CLI output:

```
[AppID 200710, ActionID 5] : LaunchApp changed task to SynchronizingCloud with ""
GameAction [AppID 200710, ActionID 5] : LaunchApp changed task to SiteLicenseSeatCheckout with ""
GameAction [AppID 200710, ActionID 5] : LaunchApp changed task to CreatingProcess with ""
GameAction [AppID 200710, ActionID 5] : LaunchApp waiting for user response to CreatingProcess ""
GameAction [AppID 200710, ActionID 5] : LaunchApp continues with user response "CreatingProcess"
Opted-in Controller Mask: 30
Game update: AppID 200710 "", ProcID 756, IP 0.0.0.0:0
>>> Adding process 756 for game ID 200710
GameAction [AppID 200710, ActionID 5] : LaunchApp changed task to WaitingGameWindow with ""
ERROR: ld.so: object '/home/przemek/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/przemek/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
/home/przemek/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Torchlight II/ModLauncher.bin.x86_64: symbol lookup error: /lib/libfontconfig.so.1: undefined symbol: FT_Done_MM_Var
GameAction [AppID 200710, ActionID 5] : LaunchApp changed task to Completed with ""
>>> Adding process 757 for game ID 200710
Game removed: AppID 200710 "", ProcID 756 
No cached sticky mapping in ActivateActionSet.
```

Side note - you rock guys. After upgrading my system I found out that Steam runtime provided in the repos for my distro causes a bit of issues. Flatpak to the rescue - majority of the games work great.

mczimmermann:
I had a similar problem with Torchlight, the solution was to delete (better move or rename) libfreetype.so in 

[...]

PR #276, opened 2018-12-10
WIP: Add trap into bash for LD_LIBRARY_PATH

### LINKED ISSUE #275: Broken game remediation through traps
Experiment with using forced initrc and traps to mitigate broken shell wrappers in games.
(eg ENV=/app/etc/shinitrc)

nanonyme:
Try to prevent games from removing Steam runtime from LD_LIBRARY_PATH. Closes #275

### DIFF
--- a/entrapment.sh
+++ b/entrapment.sh
@@ -1,0 +1,14 @@
+_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH
+function ld_library_path() {
+    if [ -n "$_OLD_LD_LIBRARY_PATH" ]
+    then
+       if [ "$_OLD_LD_LIBRARY_PATH" != "$LD_LIBRARY_PATH" ]
+       then
+           LD_LIBRARY_PATH=$_OLD_LD_LIBRARY_PATH:$LD_LIBRARY_PATH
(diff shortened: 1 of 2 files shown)


entrapment.sh
```diff
@@ -0,0 +1,11 @@
+_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH
+function ld_library_path() {
+    if [ -n "$_OLD_LD_LIBRARY_PATH" -a "$_OLD_LD_LIBRARY_PATH" != "$LD_LIBRARY_PATH" ]
+    then
+	LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$_OLD_LD_LIBRARY_PATH
```
TingPing: whitespace!

nanonyme: I really need to teach my Emacs to use spaces since I always manually use them anyway...

nanonyme:
@TingPing any opinions on the order? If I set the order the other way around, it might potentially fix #95 but maybe break some other games.

nanonyme:
Probably need to make the trap not be used for steam.sh at least

nanonyme:
Looks like this won't work, closing

PR #283, opened 2018-12-24
Update Steam to 1.0.0.59

gasinvein:
Current checksum is outdated.
It's better to use versioned archive url, as `steam_latest.tar.gz` is expected to change.

### DIFF
--- a/com.valvesoftware.Steam.json
+++ b/com.valvesoftware.Steam.json
@@ -215,8 +215,8 @@
             "sources": [
                 {
                     "type": "archive",
-                    "url": "http://repo.steampowered.com/steam/archive/precise/steam_latest.tar.gz",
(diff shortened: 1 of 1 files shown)


nanonyme:
Well, using latest has the upside that it breaks local build whenever sha256 is for wrong version so it forces you to update.

gasinvein:
Not sure if it's good, as it makes hard to reproduce build - one can't easily determine which version has that checksum.

Also, I didn't notice #282, sorry. Should I close this PR?

nanonyme:
Yeah, I've beem mitigating this through having version information in commit so it's visible through Git blame what the sha256 means.

winehub:
Sorry gasinvein. It was me. Update my comment.

PR #290, opened 2018-12-30
Add .editorconfig

gasinvein:
Created .editorconfig based on current code style. Maybe it would help avoid inconsistent code style in the future.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+root = true
+
+[*]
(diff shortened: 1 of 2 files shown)


nanonyme:
Or then rewrite JSON to YAML do inconsistent indentation an error

gasinvein:
Yeah, I'd say yaml is better, especially for manually written data.

PR #296, opened 2019-01-03
Path simplification

nanonyme:
Try to remove .var/app/com.valvesoftware.Steam from XDG vars. It should never have been there but ended there as a result of Steam Cloud fixup. The end result was crazy long paths which breaks some games.


nanonyme:
@TingPing FWIW

nanonyme:
Argh! Looks there's some symlinks inside .steam directory that prevent this fix from automatically working so user will need to delete .var/app/com.valvesoftware.Steam/.steam. I don't want to delete it in this PR because it will lbreak everyone's logins...

PR #298, opened 2019-01-04
Update summary

LeandroStanger:

### DIFF
--- a/com.valvesoftware.Steam.appdata.xml
+++ b/com.valvesoftware.Steam.appdata.xml
@@ -4,7 +4,7 @@
   <id type="desktop">com.valvesoftware.Steam.desktop</id>
   <metadata_license>CC0-1.0</metadata_license>
   <name>Steam</name>
-  <summary>Manage and play games distributed by Steam</summary>
+  <summary>Play popular games and the latest releases</summary>
   <summary xml:lang="pt_BR">Jogue jogos populares e os últimos lançamentos</summary>
   <summary xml:lang="es">Juega a los juegos populares y los últimos lanzamientos</summary>
   <description>


nanonyme:
Where's this string from?

com.valvesoftware.Steam.appdata.xml
```diff
@@ -4,7 +4,7 @@
   <id type="desktop">com.valvesoftware.Steam.desktop</id>
   <metadata_license>CC0-1.0</metadata_license>
   <name>Steam</name>
-  <summary>Manage and play games distributed by Steam</summary>
+  <summary>Play popular games and the latest releases</summary>
```
sat3ll: The previous summary seems a better fit although I understand your motivation to change it to this. The "pt_BR" and "es" languages are literal translations of your new proposal for the "english" summary. In my view it is not necessary to "harmonise"/make them all the same literal translation. Literal translations often make the phrase sound "strange". As long the intended meaning/message is conveyed, there's no problem for the translations to "diverge" literally.

sat3ll (commented):
Literal translation of "es" and "pt_BR"

com.valvesoftware.Steam.appdata.xml
nanonyme: Possibly "Play latest releases of popular games" if you want the change. But again, where's these strings from? Are they from Valve?

sat3ll: I can't find them anywhere on steam site (set to pt_BR). Git blame points the "es" and "pt_BR" translations to @LeandroStanger so I assume he translated them himself.

PR #302, opened 2019-01-08
Add hwdb/hwdata

gasinvein:
Add hwdata for pciutils and enable systemd's hwdb for usbutils, so `lspci` and `lsusb` would list device names, not just ids.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -99,10 +99,18 @@ modules:
         url: https://gitlab.freedesktop.org/libbsd/libbsd.git
         commit: 0.9.1-0-g1ca09c18f76a0ed2ef6349463c25d509a55e20ff
 
+  - name: hwdata
+    config-opts:
+      - --datarootdir=/app/share
+    sources:
+      - type: archive
+        url: https://github.com/vcrhonek/hwdata/archive/v0.312.tar.gz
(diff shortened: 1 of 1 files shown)


nanonyme:
lspci is probably going to be removed, see https://gitlab.com/freedesktop-sdk/freedesktop-sdk/issues/584

nanonyme:
I dunno, I'll need to discuss this a bit with jjardon. lspci is still pretty small but if we start talking of libsystemd dependencies, things start expanding maybe too much.

gasinvein:
Systemd is already in the bundle anyway, so why not use it's hwdb for 
libusb?

nanonyme:
Would it work if lspci was not in app? It probably doesn't make any sense to ship it twice.

gasinvein:
If pciutils (along with hwdata) would be provided by runtime, it can be safely removed from the app bundle, usbutils isn't tied to it.

gasinvein:
I've found out that hwdata is actually used by `lsusb.py`. So, once hwdata is added to the runtime and removed from the bundle, this
https://github.com/flathub/com.valvesoftware.Steam/blob/e1a0b6df220e8066f1db146e2c8a2d2a817487ed/com.valvesoftware.Steam.yml#L185-L186
should be changed to `/usr/share/hwdata`.

PR #304, opened 2019-01-09
Fix lsusb.py

gasinvein:
Seems like lsusb.py is incorrectly converted to Python 3. Add a patch (from Fedora) fixing it.
Follow-up for #302.


nanonyme:
Is there an upstream ticket about this?

gasinvein:
Yes, looks like it's already upstreamed: https://github.com/gregkh/usbutils/commit/a7c25eadbc998bf359e5b7dac03aaea8c30d2932

PR #349, opened 2019-04-16
Add support for proton extensions

### LINKED ISSUE #348: Add extensions for custom proton builds
I'm thinking it would be quite neat to add extensions to add custom proton builds made by community members.

The script would need to be modified to move the contents of say, `/app/proton/proton_tkg` or `/app/proton/proton_ge` to `~/.var/app/com.valvesoftware.Steam/.steam/root/compatibilitytools.d`. (that folder also needs to be created)

I think good candidates for those builds would be :
https://github.com/Tk-Glitch/PKGBUILDS/releases/tag/4.6.r0.g3f8edce5
https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/4.6-GE-1

Discuss if there's any issues with doing this and/or ways to improve the idea.

lionirdeadman:
Ready for review, will fix #348

### DIFF
--- a/com.valvesoftware.Steam.Proton_tkg.yml
+++ b/com.valvesoftware.Steam.Proton_tkg.yml
@@ -1,0 +1,13 @@
+id: com.valvesoftware.Steam.Proton_tkg
+runtime: com.valvesoftware.Steam
+sdk: org.freedesktop.Sdk//18.08
+modules:
+
+  - name: wine
+    buildsystem: simple
+    build-commands:
+      - "cp -r . ${FLATPAK_DEST}"
+    sources:
+      - type: archive
+        url: https://github.com/Tk-Glitch/PKGBUILDS/releases/download/4.6.r0.g3f8edce5/proton_tkg_4.6.r0.g3f8edce5.tar.gz
+        sha256: dc9d6f067ed36df0dd6ebbcfb09ddc6ba0bd5f003ca675c29289ab7ea4e5ab79
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -63,6 +63,10 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
+  com.valvesoftware.Steam:
+    subdirectories: true
+    directory: proton
+
 cleanup:
   - /lib/*.a
   - /lib/*.la
@@ -251,6 +255,12 @@ modules:
         sha256: cebf0046bfd08cf45da6bc094ae47aa39ebf4155e5ede41373b579b8f1071e7c
       - type: file
         path: com.valvesoftware.Steam.cmd.metainfo.xml
+
+  - name: proton_extensions
+    buildsystem: simple
+    build-commands:
+      - |
+        mkdir /app/proton
 
   - name: platform-bootstrap
     buildsystem: simple
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -66,6 +66,8 @@ add-extensions:
   com.valvesoftware.Steam.Proton:
     subdirectories: true
     directory: proton
+    no-autodownload: true
+    autodelete: true
 
 cleanup:
   - /lib/*.a
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -63,7 +63,7 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
-  com.valvesoftware.Steam:
+  com.valvesoftware.Steam.Proton:
     subdirectories: true
     directory: proton
 
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -30,8 +30,9 @@ def read_flatpak_info(path):
     return {
         "flatpak-version": flatpak_info.get("Instance", "flatpak-version"),
         "runtime-path": flatpak_info.get("Instance", "runtime-path"),
-        "app-extensions": flatpak_info.get("Instance", "app-extensions",
-                                           fallback=None),
+        "app-extensions": dict((s.split("=")
+                                for s in flatpak_info.get("Instance", "app-extensions",
+                                                          fallback="").split(";"))),
         "runtime-extensions": flatpak_info.get("Instance",
                                                "runtime-extensions",
                                                fallback=None),
@@ -131,12 +132,11 @@ def check_bad_filesystem_entries(entries):
                "Frequently-asked-questions#i-want-to-add-external-disk-for-steam-libraries")
         raise SystemExit(f"Please see {faq}")
 
-def check_allowed_to_run():
-    current_info = read_flatpak_info(FLATPAK_INFO)
+def check_allowed_to_run(current_info):
     current_version = current_info["flatpak-version"]
     required = "0.9.0"
     if LooseVersion(current_version) < LooseVersion(required):
-        raise SystemExit(f"Flatpak {required} or newer required")    
+        raise SystemExit(f"Flatpak {required} or newer required")
 
     check_bad_filesystem_entries(current_info["filesystems"])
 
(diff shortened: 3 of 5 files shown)


com.valvesoftware.Steam.yml
```diff
@@ -63,6 +63,10 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
+  com.valvesoftware.Steam:
```
gasinvein: I guess since this extension point is specific to Proton, it should be named accordingly.
```suggestion
  com.valvesoftware.Steam.Proton:
```
And proton build can be named like `com.valvesoftware.Steam.Proton.tkg`

gasinvein (changes requested):

steam_wrapper/steam_wrapper.py
```diff
@@ -246,6 +246,26 @@ def configure_shared_library_guard():
...
+
+    # Try to create folder if it doesn't exist
+    os.makedirs(proton_dest, exist_ok=True)
+
+    # Copy extensions if they exist
+    subfolders = [f.path for f in os.scandir("/app/proton") if f.is_dir() ]
+
+    for proton in subfolders:
+        shutil.copy(proton, proton_dest)
+
```
gasinvein: In this case actual Proton build wouldn't be updated when flatpak extension is updated.
We need either implement a timestamp check to copy the extension to users directory again if the later is outdated, or just symlink instead of copying.

com.valvesoftware.Steam.yml
```diff
@@ -63,6 +63,10 @@ add-extensions:
...
+    subdirectories: true
+    directory: proton
```
nanonyme: Please set
```
no-autodownload: true
autodelete: true
```
This should be opt-in.

steam_wrapper/steam_wrapper.py
```diff
@@ -246,6 +246,26 @@ def configure_shared_library_guard():
...
+        os.symlink(src, dst)
+
+    # Try to create folder if it doesn't exist
+    os.makedirs(proton_dest, exist_ok=True)
+
+    # Copy extensions if they exist
+    subfolders = [f.path for f in os.scandir("/app/proton") if f.is_dir() ]
+
+    for proton in subfolders:
+        shutil.copy(proton, proton_dest)
```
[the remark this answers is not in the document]
lionirdeadman: Removing stale content is probably not something we want to do since wine is bound to break certain games after updates so it's still nice to be able to let them keep the old versions.

steam_wrapper/steam_wrapper.py
lionirdeadman: symlinking is not possible due to how proton functions, it needs to be mutable. Although, not sure why it wouldn't be updated when flatpak extensions are updated?

com.valvesoftware.Steam.Proton_tkg.yml
```diff
@@ -0,0 +1,13 @@
...
+sdk: org.freedesktop.Sdk//18.08
+modules:
+
+  - name: wine
+    buildsystem: simple
+    build-commands:
+      - "cp -r . ${FLATPAK_DEST}"
+    sources:
+      - type: archive
+        url: https://github.com/Tk-Glitch/PKGBUILDS/releases/download/4.6.r0.g3f8edce5/proton_tkg_4.6.r0.g3f8edce5.tar.gz
```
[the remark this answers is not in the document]
lionirdeadman: 19.08 allows 32bit I'm guessing?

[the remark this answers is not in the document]
nanonyme: There's an in-progress Sdk extension that will allow you to cross-compile to i686 from x86_64.

lionirdeadman:
Okay, status update.

TKG's build doesn't work atm because gnutar can't unpack it.
GE's build works in that regard.

Here's the bad part :
1. I realized keeping old copies is not really something I can do due to the structure.
2. Something makes the copy loop itself with GE's build which seems to be related to Z: being Home.
3. I don't know how we should check to see the extension matches compatibility.d's since it's mutable and unpacked so the other option is basically checking folder names or maybe I'm missing something. This is needed though because copying and deleting 600MB every time Steam starts is.. noticeable.

com.valvesoftware.Steam.Proton_tkg.yml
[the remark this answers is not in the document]
lionirdeadman: I'll mark this as resolved since 19.08 is not out yet. Is that fine?

lionirdeadman:
Okay so now :

- Things are checked by commit and then replaced if commit changed
- TKG builds work
- The copy issue is fixed

Everything works now but I think we should make steam and the copy start at the same time so the copy doesn't seem to have an effect on the start time

lionirdeadman:
And with this, I think everything is good to be reviewed and shipped.

steam_wrapper/steam_wrapper.py
```diff
@@ -246,10 +248,38 @@ def configure_shared_library_guard():
...
+        proton_ext_commit_file = proton_ext_dest / '.extension-commit'
+
+        if proton_ext_dest.exists():
+            if proton_ext_commit_file.exists():
+                with proton_ext_commit_file.open() as fp:
+                    if fp.read() == proton_ext_commit_hash:
+                        continue
+
+        print(f"Copying extension {proton_ext}")
+        if proton_ext_dest.exists():
```
nanonyme: If you want to avoid copying every time, just write a manifest during build time (sha256 sums of every file) and check if manifest has changed during runtime. It's unnecessary to actually verify hashes if you want the directories to be largely mutable.

lionirdeadman: This has been solved already above by using flatpak commit hashes right above thanks to @refi64's help though?.. It doesn't actually check files themselves, it just checks if the file hash is the same as when it was first copied. This allows for both mutability, speed and still keeping the copy in tandem with the extension.

nanonyme: I think the hash resets every time there's any change within this app but could be wrong.

nanonyme: I guess this is fine. If we hit issues, let's improve later.

steam_wrapper/steam_wrapper.py
```diff
@@ -258,4 +288,8 @@ def main(steam_binary=STEAM_PATH):
     timezone_workaround()
     configure_shared_library_guard()
     enable_discord_rpc()
-    os.execve(steam_binary, [steam_binary] + sys.argv[1:], os.environ)
+    n = os.fork()
+    if n == 0:
```
nanonyme: I don't like this. If you want to keep running, remove execve and instead call steam using subprocess. Fork without exec is not really safe.

lionirdeadman: My latest commit now this does, marking as resolved.

steam_wrapper/steam_wrapper.py
refi64: Some explanation behind this:

It basically came down to that checking individual sha sums of these files often ended up slowing things down more than they helped, and I doubt there's going to be rebuilds unless something actually changed. Manual sha comparisons also introduce the potential for random stale files if something *does* change, and the code would be a bit more complex than expected since Proton distributions often have default wineprefixes with e.g. recursive symlinks.

TL;DR: not really worth it for now...

nanonyme: Afaik Flathub always recompiles entire app whenever *anything* changes. Do you think it makes sense eg systemd version bump results in wiping and recreating the Proton data? I'm not entirely sure if this happens with app extensions but would expect so.

nanonyme: Also as said, build-time hashing does not really slow things down for user. In runtime you can just copy over manifest and go no-op if manifest under /app and under deploy dir are identical (without actually verifying files).

refi64: Whenever...anything changes? The commit of the extension is checked, and as the original app is used as a runtime, the only times the commit should change would be when the extension itself is rebuilt.

nanonyme: You're confused app extension is built as part of the application when done the way described in this PR. Any app change results in the extension to be rebuilt too afaik.

gasinvein: @nanonyme I assumed the the extension's manifests are here just for example, and later they will be moved to separate repositories. Are they going to be maintained here, as a part of the application?

nanonyme: I assumed this was more or less the final plan. I don't know any context beyond thos PR and related ticket.

nanonyme:
I don't think this is a viable long-term solution since first of all to avoid forward ABI issues we don't want to use precompiled binaries from this project in the long run and the building relies too heavily on pacman. The pacman dependency needs to be refactored out.

lionirdeadman:
> I don't think this is a viable long-term solution since first of all to avoid forward ABI issues we don't want to use precompiled binaries from this project in the long run and the building relies too heavily on pacman. The pacman dependency needs to be refactored out.

This most likely won't happen unless one of you want to refactor TKG's code to do this. I don't have the experience to do this and it's my understanding that he doesn't have plans to do so either from what I know. Would it be possible to build pacman to build TKG's code instead?

nanonyme:
Not really. Pacman assumes running on top of Arch Linux. We don't want to build this on Arch but freedesktop-sdk to avoid compatibility issues.

lionirdeadman:
I don't think it does, we have [pacman for Fedora.](https://apps.fedoraproject.org/packages/pacman)

gasinvein:
Pacman for Fedora is primarily intended for creating chroots, as is dpkg/debhelper for Fedora. It still uses Arch Linux repositories, whose packages are still incompatible with your Fedora systemd.

lionirdeadman:
Well, alright. I'm not sure how to make a wrapper for TKG's PKGBuild, if any of you know how / want to do it, feel free.

gasinvein:
I've [flatpak'ed pacman](/gasinvein/flatpak-apps-staging/tree/org.freedesktop.Sdk.Extension.pacman/master) as a FD.o sdk extension. It can at least run `makepkg --nodeps` so the `PKGBUILD`'s can be reused, but it's not very helpful, though.

gasinvein:
Actually, Steam compatibility tools are not limited to Proton. For example, there is a [Dosbox tool](https://github.com/dreamer/steam-dos). Maybe it's better to name the extension point more generically, e.g. `com.valvesoftware.Steam.CompatibilityTool`?

lionirdeadman:
Yes that might be worthwhile but since I'm not sure how to go about doing custom proton builds since we don't want to use prebuilt binaries, I think I'll leave it to someone else.

gasinvein:
I think it's ok to add the extension point and tweak the wrapper, while dropping the Proton manifests, since they have to be submitted to Flathub separately anyway. But it's up to @nanonyme to decide.

nanonyme:
That sounds fine to me. That also postpones discussions about redistributability to a followup.

gasinvein:
I guess this is obsolete now?

nanonyme:
Please open new PR against beta if further changes desired.

PR #350, opened 2019-04-29
Add reference to old desktop ID

AsciiWolf:
See [this comment](https://github.com/flathub/net.sourceforge.Fillets/issues/4#issue-436137709).

### DIFF
--- a/com.valvesoftware.Steam.appdata.xml
+++ b/com.valvesoftware.Steam.appdata.xml
@@ -3,6 +3,9 @@
 <application>
   <id type="desktop">com.valvesoftware.Steam.desktop</id>
   <metadata_license>CC0-1.0</metadata_license>
+  <provides>
+    <id>steam.desktop</id>
+  </provides>
   <name>Steam</name>
   <summary>Manage and play games distributed by Steam</summary>
   <summary xml:lang="pt_BR">Jogue jogos populares e os últimos lançamentos</summary>


nanonyme:
I'm not sure this is ok. Valve doesn't officially support this application and their stand is this is fine as long as people won't confuse it with official Flatpak.

AsciiWolf:
Hmm, I am not sure if this is an issue. Flathub is not enabled by default on most distributions and Flatpak is not even installed by default on Ubuntu. Also, almost every distribution has its own Steam package. Are those supported?

Anyway, with the correct old desktop ID, Steam can be displayed as one entry in GNOME Software (since 3.32) and user can easily select which repository source they want to choose. Without the old desktop ID, it is displayed as multiple entries and it is actually confusing for the user.

@hughsie What is your opinion about this?

hughsie:
> What is your opinion about this?

It needs the patch as suggested by @AsciiWolf -- **it's the same app** -- any developer information about who packaged it, or who supplied the binaries is component information that is shown in the software center as extra information.

gasinvein:
Isn't `rename-desktop-file` enough? flatpak-builder adds a `X-Flatpak-RenamedFrom` entry to the .desktop file, which is read by gnome-software to identify same app.

AsciiWolf:
@gasinvein Hmm, I am not sure. Maybe @hughsie will know more. :-)

gasinvein:
@AsciiWolf well, at least for me gnome-software already correctly identifies Steam from different sources as being the same app:
Screenshot 2019-05-03 15-10-45

AsciiWolf:
Interesting.

AsciiWolf:
@hughsie @alexlarsson Is the `rename-desktop-file` enough or should there be the old desktop ID added to the AppData file?

AsciiWolf:
@gasinvein I just tried it on Fedora 30. You're right, it works fine even without the old desktop ID reference in AppData.

PR #363, opened 2019-05-26
Implement lsb_release shim

nanonyme:
Write a shim that just reads from /etc/os-release

### DIFF
--- a/resources/lsb_release
+++ b/resources/lsb_release
@@ -1,0 +1,45 @@
+#!/bin/bash
+. /etc/os-release
+
+identity() {
+    printf "Distributor ID:\t$NAME\n"
+}
+
+description() {
+    printf "Description:\t$PRETTY_NAME\n"
+}
+
+release() {
+    printf "Release:\t$VERSION_ID\n"
+}
+
+codename() {
+    printf "Codename:\tn/a\n"
+}
+
+for argument in "$@"
+do
+    case $argument in
+        -h|--help)
+            echo This is a shim that reads from /etc/os-release
+            ;;
+        -i|--id)
+            identity
+            ;;
+        -d|--description)
+            description
+            ;;
+        -r|--release)
+            release
+            ;;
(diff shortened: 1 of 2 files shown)


resources/lsb_release
```diff
@@ -0,0 +1,53 @@
...
+description() {
+    printf "Description:\t $PRETTY_NAME\n"
+}
+
+release() {
+    printf "Release:\t $VERSION_ID\n"
+}
+
+supported_modules() {
+    echo LSB version:
```
gasinvein: Ubuntus `lsb_release` doesn't print the this line at all (_not_ an empty list) in case of no LSB modules are installed (which is default).

nanonyme: Ah, great. I'll omit. What does it do if you explicitly request -v? Nothing?

gasinvein: Prints a message to stderr saying that no LSB modules are installed.

gasinvein:
Maybe make this shim more universal and put it in a separate repo? It can be useful for others.
Sadly, there are some apps around relying to `lsb_release` command.

resources/lsb_release
nanonyme: Ack. I guess it's not really that important here.

nanonyme:
Maybe, I just want to first know if this fixes Steam survey.

resources/lsb_release
```diff
@@ -0,0 +1,45 @@
...
+	-c|--codename)
+	    codename
+	    ;;
+	-a|--all)
+	    identity
+	    description
+	    release
+	    codename
+	    ;;
+esac
```
gasinvein:
```suggestion
esac
    esac
```

gasinvein: Tab/spaces mixtures here.

nanonyme: Fixed.

nanonyme:
Tested, seems to be enough for Steam

PR #365, opened 2019-05-27
Put libbfd in correct directory, fixing addr2line

gasinvein:
I've suddenly notices that `libbfd-2.31.1.so` is copied into `/app/lib/x86_64-linux-gnu`, which is not under ld library path. Thus, `addr2line` isn't working.


nanonyme:
I'm really confused but sounds right.

PR #366, opened 2019-05-29
Change path to Discord-RP socket

gasinvein:
Discord-RP socket path was changed in flathub/com.discordapp.Discord#68, so change it here accordingly.


nanonyme:
I'm a bit worried if we leave leftovers but the change seems sensible.

gasinvein:
Leftovers?

PR #368, opened 2019-06-08
Add FAudio

gasinvein:
FAudio with wma support (via ffmpeg) is needed for some Windows games, playable with Proton.
This currently would work only for 64-bit games, but I hope to fix it once FD.o SDK 19.08 become available.

FAudio libs shipped with Proton have to be ~~manually deleted~~ blocked via shared-library-guard to make this work, see https://github.com/FNA-XNA/FAudio/wiki/FAudio-for-Proton for details.


nanonyme:
Imo better approach is here https://gitlab.com/freedesktop-sdk/freedesktop-sdk/merge_requests/918

nanonyme:
I don't like the idea of writing over Steam files in general though, we should probably see if shared-library-guard can be used here.

gasinvein:
>Imo better approach is here

Sadly it won't work. ffmpeg support in FAudio has to be enable at compile-time, which is not the case for the builds bundled with Proton.

>I don't like the idea of writing over Steam files in general though, we should probably see if shared-library-guard can be used here.

That would be definitely better, if possible. Yet I'm totally unaware of how the shared-library-guard works.

nanonyme:
It's basically just a libc audit library.

gasinvein:
Does shared-library-guard work for shared libraries, or only for executables?
It's `Proton\ */dist/lib64/wine/xaudio2_*.dll.so` linked with `libFAudio.so`.

nanonyme:
The tool is likely to get refactored a lot with 19.08 as it will then move to be built in Steam app and can have more specific logic. For actual answer all libraries get loaded by executables. In the case of Proton tha process is most probably wine. The audit API supports going beyond blocking, you can also directly return path to library.

gasinvein:
AFAIK wine `dlopen`s its libraries when they are needed. Is it an obstacle?

I've tried adding this to `freedesktop.ld.so.blockedlist`:
```
Proton\ 4.2/dist/bin/wine Proton\ 4.2/dist/lib64/libFAudio.so.0
Proton\ 4.2/dist/bin/wine64 Proton\ 4.2/dist/lib64/libFAudio.so.0
Proton\ 4.2/dist/bin/wine64-preloader Proton\ 4.2/dist/lib64/libFAudio.so.0
Proton\ 4.2/dist/bin/wine-preloader Proton\ 4.2/dist/lib64/libFAudio.so.0
Proton\ 4.2/dist/bin/wineserver Proton\ 4.2/dist/lib64/libFAudio.so.0
```
and it doesn't seem to have any effect.

gasinvein:
So I've forgot to enable shared-library-guard with the env var. It actually works and blocks the proton-provided fauidio just fine.
Does the shared-library-guard support wildcards in it blocklist?

nanonyme:
No, wildcards not supported. Just matches at end of path so semantically beginning of path can be considered wildcard.

gasinvein:
I'm asking because FAudio soversion may vary for a single Proton version, so without wildcards the blocklist would probably require frequent updates.

nanonyme:
I see. We can change tweak how it works although it's a pure-C codebase so it might be a bit painful to do complex stuff with. I would prefer not to change it much for now since until it moves to app we have quite long integration cycles.

nanonyme:
Source code lives here https://gitlab.com/freedesktop-sdk/shared-library-guard

com.valvesoftware.Steam.yml
```diff
@@ -105,6 +105,8 @@ modules:
 
   - "modules/discord.yml"
 
+  - "modules/faudio.yml"
```
nanonyme: I think it would be more obvious what the purpose is if the module was called proton.

gasinvein: If you say so.

PR #374, opened 2019-06-21
Add support for compatibility tools extensions

gasinvein:
This is changeset from #349 with merge conflicts resolved and "Proton" renamed to "CompatibilityTool".


steam_wrapper/steam_wrapper.py
```diff
@@ -255,4 +292,5 @@ def main(steam_binary=STEAM_PATH):
     timezone_workaround()
     configure_shared_library_guard()
     enable_discord_rpc()
-    os.execve(steam_binary, [steam_binary] + sys.argv[1:], os.environ)
+    p = subprocess.Popen([steam_binary] + sys.argv[1:])
```
gasinvein: @nanonyme @refi64 If we can avoid IO-intensive copying of the whole extension, maybe we should switch back to using `os.execve`?

nanonyme (commented):
If we can, we should use simple execv.

steam_wrapper/steam_wrapper.py
```diff
@@ -243,10 +245,45 @@ def configure_shared_library_guard():
         if mode > 1:
             os.environ["LD_BIND_NOW"] = "1"
 
+def setup_compat_tool_extensions(current_info):
+    compat_tool_dest = Path('.local/share/Steam/compatibilitytools.d')
```
gasinvein: I wonder, why this path is relative?

nanonyme:
Would you mind re-targeting against beta branch?

gasinvein:
Should I open another PR, or just change base branch for this one?

nanonyme:
I was intending to add all commits from beta to master by rebasing once 19.08 goes live so not really preferring to merge things into master anymore since will make life harder in one month.

nanonyme:
I was intending to add all commits from beta to master by rebasing once 19.08 goes live so not really preferring to merge things into master anymore since will make life harder in one month.

nanonyme:
Rebased on top of beta, will merge once build passes.

steam_wrapper/steam_wrapper.py
nanonyme: Doesn't matter, the script initially chdirs to home to have consistent paths.

com.valvesoftware.Steam.yml
```diff
@@ -75,6 +75,12 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
+  com.valvesoftware.Steam.CompatibilityTool:
+    subdirectories: true
+    directory: compatibilitytools.d
```
gasinvein: Multiple 3rd-partiy compatibility tools install themselves to `$prefix/share/steam/compatibilitytools.d`. Maybe we should do the same?

nanonyme: I haven't actually tried these so I'm going to trust others on this for now. I guess that means when you create an extension, you would drop files into the root directory.

gasinvein: Well, I guess the mountpoint could be changed later anyway.

Issue #105, opened 2018-07-10
BlackMesa - libtcmalloc_minimal.so.4 not found


alexryndin:
I'm trying to run blackmesa on Arch Linux. Steam installed and run using flatpak as following:
``` 
 flatpak --user remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
 flatpak --user install flathub com.valvesoftware.Steam
 flatpak run com.valvesoftware.Steam
```
Steam starts with no problem, but when I try to lauch BM, I get following error:
```
$HOME/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Black Mesa/bin:$HOME/.var/app/com.valvesoftware.Stea
m/data/Steam/ubuntu12_32/steam-runtime/pinned_libs_32:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/ste
am-runtime/pinned_libs_64:/app/lib/32bit/lib:/app/lib:/usr/lib/ffmpeg/lib:/lib:/lib64:$HOME/.var/app/com.valvesoftware.
Steam/data/Steam/ubuntu12_32/steam-runtime/i386/lib/i386-linux-gnu:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ub
untu12_32/steam-runtime/i386/lib:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/steam-runtime/i386/usr/l
ib/i386-linux-gnu:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/steam-runtime/i386/usr/lib:$HOME/.var/a
pp/com.valvesoftware.Steam/data/Steam/ubuntu12_32/steam-runtime/amd64/lib/x86_64-linux-gnu:$HOME/.var/app/com.valvesoft
ware.Steam/data/Steam/ubuntu12_32/steam-runtime/amd64/lib:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32
/steam-runtime/amd64/usr/lib/x86_64-linux-gnu:$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/steam-runti
me/amd64/usr/lib:
ERROR: ld.so: object '$HOME/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRE
[...]

PR #387, opened 2019-07-16
Add google-perftools as 32bit

### LINKED ISSUE #69: "Half-Life 2: Update" does not launch because of missing shared library
"Half-Life 2: Update", a linux-compatible-game, cannot launch. Here is the relevant error message from the terminal:

~/.var/app/com.valvesoftware.Steam/data/Steam/steamapps/common/Half-Life 2 Update/hl2_linux: error while loading shared libraries: libtcmalloc_minimal.so.4: cannot open shared object file: No such file or directory

nanonyme:
Fixes #69 #105

### DIFF
--- a/modules/compat-libraries-32bit.yml
+++ b/modules/compat-libraries-32bit.yml
@@ -19,4 +19,5 @@ modules:
     sources:
       - type: git
         url: https://github.com/gperftools/gperftools.git
-        commit: perftools-1.6-0-gfd52267cb91a42f839c7be6a348e014cfdbcaca9
+        commit: 0b87a55895a184bcc529b4ce13c46c0411182b85
+        tag: google-perftools-1.8.3
--- a/modules/compat-libraries-32bit.yml
+++ b/modules/compat-libraries-32bit.yml
@@ -16,6 +16,8 @@ modules:
     build-options:
       arch:
         x86_64: *compat_i386_opts
+      cflags: -fpermissive -Wno-error=narrowing
+      cxxflags: -fpermissive -Wno-error=narrowing
     sources:
       - type: git
         url: https://github.com/gperftools/gperftools.git
--- a/modules/compat-libraries-32bit.yml
+++ b/modules/compat-libraries-32bit.yml
@@ -9,8 +9,9 @@ x-compat-i386-opts: &compat_i386_opts
 
 name: compat-libraries-32bit
 buildsystem: simple
-commands: []
+build-commands: []
 modules:
+
   - name: google-perfools
     buildsystem: autotools
     build-options:
--- a/modules/compat-libraries-32bit.yml
+++ b/modules/compat-libraries-32bit.yml
@@ -4,7 +4,7 @@ x-compat-i386-opts: &compat_i386_opts
   prepend-path: /usr/lib/sdk/toolchain-i386/bin
   env:
     CC: i686-unknown-linux-gnu-gcc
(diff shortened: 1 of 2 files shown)


modules/compat-libraries-32bit.yml
```diff
@@ -0,0 +1,22 @@
+x-compat-i386-opts: &compat_i386_opts
+  prepend-pkg-config-path: /app/lib32/pkgconfig:/usr/lib/i386-linux-gnu/pkgconfig
+  ldflags: -L/app/lib32
+  prepend-path: /usr/lib/sdk/toolchain-i386/bin
+  env:
+    CC: i686-unknown-linux-gnu-gcc
+    CXX: i686-unknown-linu-gnu-gcc
```
gasinvein:
```suggestion
    CXX: i686-unknown-linux-gnu-g++
```

modules/compat-libraries-32bit.yml
```diff
@@ -0,0 +1,22 @@
...
+modules:
+  - name: google-perfools
+    buildsystem: autotools
+    build-options:
+      arch:
+        x86_64: *compat_i386_opts
+    sources:
+      - type: git
+        url: https://github.com/gperftools/gperftools.git
+        commit: perftools-1.6-0-gfd52267cb91a42f839c7be6a348e014cfdbcaca9
```
gasinvein: Seems like gperftools 1.6 won't build with newer glibc: gperftools/gperftools#350
```suggestion
        commit: 0b87a55895a184bcc529b4ce13c46c0411182b85
        tag: google-perftools-1.8.3
```

nanonyme: Is this just head or some specific commit?

gasinvein: It's the commit tagged as version 1.8.3.

gasinvein:
@nanonyme Maybe rename this module to something more meaningful, e.g. `game-libraries-32bit.yml`? Other module names clearly tell what they are needed for, but not this one.

modules/compat-libraries-32bit.yml
```diff
@@ -0,0 +1,26 @@
+x-compat-i386-opts: &compat_i386_opts
+  prepend-pkg-config-path: /app/lib32/pkgconfig:/usr/lib/i386-linux-gnu/pkgconfig
+  ldflags: -L/app/lib32
+  prepend-path: /usr/lib/sdk/toolchain-i386/bin
+  env:
+    CC: i686-unknown-linux-gnu-gcc
+    CXX: i686-unknown-linux-gnu-g++
+    libdir: /app/lib32
```
gasinvein: `libdir` here is supposedly not an env var, but a build options, which tells the build system where to install libraries, similarly to `prefix`.
```suggestion
  libdir: /app/lib32
```

nanonyme:
I'm going to leave renaming for after this since these further work depending on this MR.

gasinvein:
@nanonyme  Are you planning to put _all_ 32bit libraries into the `compat-libraries-32bit.yml` module?

nanonyme:
@gasinvein sadly probably yes. Otherwise there will be *massive* repetition of the 32bit definitions since cross-include anchors aren't a thing.

gasinvein:
@nanonyme I believe repeating 32bit definitions is less evil than repeating sources and config-opts. The later could cause inconsistency between 64 and 32-bit builds of the same library.

PR #388, opened 2019-07-21
Update some modules to newest versions

LeandroStanger:


nanonyme:
Please, do version bumps to beta branch. Master is more or less frozen until 19.08 runtime release unless things are badly broken.

PR #389, opened 2019-07-21
Update some modules to newest versions

LeandroStanger:

### DIFF
--- a/modules/proton.yml
+++ b/modules/proton.yml
@@ -4,8 +4,8 @@ config-opts:
   - -DFFMPEG=ON
 sources:
   - type: archive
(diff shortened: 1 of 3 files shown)


modules/proton.yml
```diff
@@ -4,8 +4,8 @@ config-opts:
   - -DFFMPEG=ON
 sources:
   - type: archive
-    url: https://github.com/FNA-XNA/FAudio/archive/19.06.07.tar.gz
-    sha256: 3a062ac74e787d2c50a5d178e9021ca4b2d4dddbbea6d36b82c767eb7f957128
+    url: https://github.com/FNA-XNA/FAudio/archive/19.07.tar.gz
```
gasinvein: Latest Proton uses FAudio `19.06.07`. We should match it.

PR #397, opened 2019-07-31
Update FAudio to 0.19.07 for Proton 4.11

gasinvein:
The new Proton 4.11 shipped with FAudio 0.19.07.
I'm still not sure what we should put in the blocklist - the actual library (`libFAudio.so.0.19.07`), the major-versioned symlink which wine is linked with (`libFAudio.so.0`), or both?

### DIFF
--- a/modules/proton.yml
+++ b/modules/proton.yml
@@ -4,8 +4,8 @@ config-opts:
   - -DFFMPEG=ON
 sources:
   - type: archive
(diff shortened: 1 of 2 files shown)


nanonyme:
@valentindavid this makes me wonder if we should just change library matching to use simple substring match.

gasinvein:
@nanonyme Please merge this into beta, too.

PR #407, opened 2019-08-03
Disable systemd-resolved in systemd build

Erick555:
systemd-resolved shouldn't be needed. It's also unlikely to work at all.

### DIFF
--- a/modules/controllers.yml
+++ b/modules/controllers.yml
@@ -45,6 +45,9 @@ modules:
       - -Dbashcompletiondir=no
       - -Dzshcompletiondir=no
       - -Dresolve=false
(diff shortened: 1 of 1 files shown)


nanonyme:
Please, active development into beta at this point. It has been refactored quite a lot so this is duplication of work.

gasinvein:
Those systemd parts are probably being removed anyway:
https://github.com/flathub/com.valvesoftware.Steam/blob/f246a93945abf5e61cdd571fab53d908e8618fab/modules/controllers.yml#L51-L54

Erick555:
@nanonyme done in https://github.com/flathub/com.valvesoftware.Steam/pull/409

PR #421, opened 2019-08-23
Reintroduce shared-library-guard

nanonyme:
It will now live inside app instead of runtime.


nanonyme:
Ping @valentindavid now finally starting this.

Issue #419, opened 2019-08-20
Proton-i, Proton-GE and Proton-tkg as an extension for Steam


LeandroStanger:
https://github.com/imaami/Proton/releases
https://github.com/GloriousEggroll/proton-ge-custom/releases
https://github.com/Tk-Glitch/PKGBUILDS/tree/master/proton-tkg

LeandroStanger:
@imaami

LeandroStanger:
@GloriousEggroll

LeandroStanger:
@Tk-Glitch

gasinvein:
This is not going to be easy, unless we just repackage upstream-provided binaries (that I believe isn't something we want to do).
Anyway, [work in progress](https://github.com/gasinvein/flatpak-apps-staging/tree/com.valvesoftware.Steam.CompatibilityTool.Proton-i/master).

gasinvein:
I've created a standalone repo for the Proton flatpak https://github.com/gasinvein/proton-flatpak
Currently the goal is to build a fully-functional vanilla Proton. Once it's done, Proton forks like proton-ge could be based on it.

gasinvein:
I think flatpak'ing Proton is mostly done, I'm using it for some time now to actually play games.
Currently only fonts are missing, but Proton is generally functional without them. Also I didn't test VR sine I don't have any VR devices.
From this point, the manifest can be pretty easily adopted for other Proton flavors (e.g. Proton-i, Proton-GE), so if anybody interested in doing this - feel free to fork the [repo](https://github.com/gasinvein/proton-flatpak).

gasinvein:
@nanonyme What's your opinion on this? Should we have native Proton-GE build on Flathub?

nanonyme:
I have no idea. Depends on how widely used it is. The only catch is I can only provide very limited support for the extensions since I don't use them myself.

gasinvein:
An what about vanilla Proton? I see two advantages of having native flatpak'ed Proton:
- It's linked with the runtime, so no ABI breakes and also no FAudio-ffmpeg-related problems
- It can have more frequent DXVK updates

The obvious downside is the maintenance effort - not sure how large it would be.

gasinvein:
[...]

Issue #394, opened 2019-07-27
Build steam-dos as an extension for Steam


LeandroStanger:
https://github.com/dreamer/steam-dos

LeandroStanger:
@dreamer

dreamer:
Hello! :) I think some work on this was already started by @gasinvein, but I don't know if it is being tracked in some issue already. Also, I think we are waiting for ValveSoftware/steam-for-linux#6310 to be released (although it's not a blocker for flathub extension if I understand correctly).

gasinvein:
Yeah I have a working manifest for building steam-dos extension [here](https://github.com/gasinvein/flatpak-apps-staging/tree/com.valvesoftware.Steam.CompatibilityTool.steam-dos/master).
ValveSoftware/steam-for-linux#6310 isn't a blocker indeed, but the extension support is added only on the beta branch of this Steam flatpak, so the steam-dos extension would be only available from flathub-beta, too.
Also I know @dreamer is planning to rename steam-dos projects, and this actually is kind of blocker since changing flatpak IDs may be tricky.

dreamer:
Yes, plan is to rename steam-dos project to Boxtron - hopefully by the end of next week (but the date is not set in stone).

In the parallel [Luxtorpeda](https://github.com/dreamer/Luxtorpeda) is being developed - it's somewhat functional, name is final, but no official release has happened yet. I think Luxtorpeda might be easier to package as an extension, as it is single Rust binary with no special dependencies.

gasinvein:
@dreamer If the name Boxtron if already final, we can just preventively give the extension ID of `com.valvesoftware.Steam.CompatibilityTool.Boxtron`, and change other things like the project name in appdata when it's actually renamed.

dreamer:
Yes, the name Boxtron is final - I am holding off with the rename because I need to test beforehand if GitHub will correctly perform redirects to the new name in all contexts (wiki links, forks, Travis, etc). I think this extension ID is good :)

gasinvein:
@nanonyme Are there any plans to merge the extension support into stable branch?

gasinvein:
Manifest with the new ID is now [here](https://github.com/gasinvein/flatpak-apps-staging/commits/com.valvesoftware.Steam.CompatibilityTool.Boxtron/master). Can confirm that it keeps working.

nanonyme:
[...]

PR #422, opened 2019-08-23
Add some dependencies for building compatibility tools

gasinvein:
Since we have the `.CompatibilityTool.` extension point, someone may want to actually build some compatibility tools for Steam. Probably we should keep most commonly-used dependencies for them, including development files.
Needed for #419 and #394.


com.valvesoftware.Steam.yml
```diff
@@ -89,12 +89,6 @@ cleanup:
   - /lib/*.la
   - /lib32/*.a
   - /lib32/*.la
-  - /include
-  - /lib/pkgconfig
-  - /lib32/pkgconfig
```
nanonyme: Hm, why are these development files needed?

gasinvein: For building compatibility tools as extensions. We don't need development files from all modules, though.

modules.yml
```diff
@@ -138,6 +92,21 @@ modules:
...
 # -- end faudio --
 
+# -- vkd3d --
+
+  - name: SPIRV-Headers
+    buildsystem: cmake-ninja
+    sources:
+      - type: archive
+        url: "https://github.com/KhronosGroup/SPIRV-Headers/archive/1.4.1.tar.gz"
+        sha256: a244f0629f75eb450e090cd773d30e22367cb231e964c7492588eb9000201fd1
```
nanonyme: I thought these come through Sdk.

com.valvesoftware.Steam.yml
nanonyme: Hmm, okay. It makes this entire app extension thing quite suboptimal. I wonder if I could create app Sdk extension that could be used....

modules.yml
gasinvein: Maybe the SDK should include these indeed, but there is no `/usr/include/spirv`.

com.valvesoftware.Steam.yml
gasinvein: It would be nice if flatpak-builder could produce some kind of SDK-like bundle of the app with development files kept in place.

nanonyme: I was mostly thinking if I could roll an app extension in this manifest and then you could map that as Sdk extension when building your app extension.

modules.yml
nanonyme: Looks like they might not be getting shipped. There's just this thing that you need to be really careful combining different projects for end result to be supported. Not all version combinations are okay. https://gitlab.com/freedesktop-sdk/freedesktop-sdk/blob/master/elements/components/spirv-tools.bst#L44

com.valvesoftware.Steam.yml
nanonyme: Maybe we need to do that as follow-up. This is atm blocking things.

PR #435, opened 2019-09-12
Fix blockedlist filename

valentindavid:
That did not work of course...


valentindavid:
Maybe we should not name it after freedesktop-sdk. Maybe `/app/etc/ld.so.blockedlist` is fine enough.

PR #438, opened 2019-09-13
Update FAudio to v19.09

LeandroStanger:
The new Proton 4.11-4 shipped with FAudio 19.09.

PR #439, opened 2019-09-15
Generate Proton blocklist at build time

gasinvein:
This adds a simple script that generates ld.so.blockedlist part for Proton. This way, we only need to define which Proton versions comes with which FAudio versions, the script will append necessary lines to the blocklist at build time.
Supersedes and closes #438.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -174,6 +174,7 @@ modules:
         cp /usr/lib/x86_64-linux-gnu/libbfd-*.so /app/lib/
         install -Dm644 -t /app/etc ld.so.conf
         install -Dm644 -t /app/etc freedesktop-sdk.ld.so.blockedlist
+        ./gen-proton-blocklist.sh >> /app/etc/freedesktop-sdk.ld.so.blockedlist
         install -Dm744 -t /app/bin lsb_release
         mkdir /app/compatibilitytools.d
         mkdir /app/links
@@ -184,5 +185,7 @@ modules:
         path: resources/ld.so.conf
       - type: file
         path: resources/freedesktop-sdk.ld.so.blockedlist
+      - type: file
+        path: resources/gen-proton-blocklist.sh
       - type: file
         path: resources/lsb_release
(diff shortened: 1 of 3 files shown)


nanonyme:
Nice idea but I don't think this can work during build-time. Proton only exists duriny runtime.

nanonyme:
Good stuff. Similar approach allows splitting game libs to separate files even if they're not dynamically generated. This keeps list more readable.

gasinvein:
This just simplifies changes need to be made for updating the blocklist on Proton update.
I.e. instead of adding
```
Proton\ 4.11/dist/bin/wine64 Proton\ 4.11/dist/lib64/libFAudio.so.19.08
Proton\ 4.11/dist/bin/wine64-preloader Proton\ 4.11/dist/lib64/libFAudio.so.19.08
Proton\ 4.11/dist/bin/wine Proton\ 4.11/dist/lib/libFAudio.so.19.08
Proton\ 4.11/dist/bin/wine-preloader Proton\ 4.11/dist/lib/libFAudio.so.19.08
Proton\ 4.11/dist/bin/wine64 Proton\ 4.11/dist/lib64/libFAudio.so.19.09
Proton\ 4.11/dist/bin/wine64-preloader Proton\ 4.11/dist/lib64/libFAudio.so.19.09
Proton\ 4.11/dist/bin/wine Proton\ 4.11/dist/lib/libFAudio.so.19.09
Proton\ 4.11/dist/bin/wine-preloader Proton\ 4.11/dist/lib/libFAudio.so.19.09
```
to `resources/freedesktop-sdk.ld.so.blockedlist`, we just need to add
```bash
PROTON_BLOCK_LIBS["Proton 4.11"]="libFAudio.so.19.08 libFAudio.so.19.09"
```
to `resources/gen-proton-blocklist.sh`.

nanonyme:
Please fix target branch to beta though and rebase. All changes should go through beta.

nanonyme:
Yes, I understood that.  Your realization that freedesktop-sdk.ld.so.blockedlist does not need to be a single static file can lead to further cleanups though.

gasinvein:
Hmm. Maybe we could implement some kind of abstraction layer, e.g. a manifest-based blocklist generator?
Thinking about something like this for a high-level manifest:
```yaml
entries:
  - basedir: Proton 4.11/dist
    blocklists:
      - binaries:
          - path: bin/wine
            libdir: lib
          - path: bin/wine-preloader
            libdir: lib
          - path: bin/wine64
            libdir: lib64
          - path: bin/wine64-preloader
            libdir: lib64
        libraries:
          - libFAudio.so.19.08
          - libFAudio.so.19.09

  - basedir: Torchlight II
    blocklists:
      - binaries:
          - path: ModLauncher.bin.x86_64
            libdir: lib64
          - path: ModLauncher.bin.x86
            libdir: lib
        libraries:
          - libfontconfig.so.1
          - libSDL2-2.0.so.0
```

nanonyme:
Maybe. What I was thinking of was about catting stuff together since the single file is most likely going to get quite large. That kind of abstraction layer might make things less effor-prone as well.

nanonyme:
I think I would prefer to make the converter be part of shared-library-guard itself though instead of hosting it here if we go there.

gasinvein:
I'll try to create a POC for this tomorrow. Is the given format self-descriptive enough?

nanonyme:
Created https://gitlab.com/freedesktop-sdk/shared-library-guard/issues/6 ; the format sounds fine to me.

Issue #485, opened 2019-10-17
steam-wrapper breaks Kindergarten (a Mono game)


nightuser:
**Game information**
Kindergarten, possibly other Mono games (haven't tested them yet)

**Distribution name and version where applicable**
ArchLinux

**Flatpak info**
```
[~]$ flatpak --version 
Flatpak 1.5.0
[~]$ flatpak --gl-drivers 
default
host
```

**Problem description**
The game ran fine a week ago. Now it doesn't work when using `steam-wrapper`. If I simply run it with `steam` everything works. Hasn't tested other Mono games yet, will report after testing.

**Does this issue reproduce with native Steam**
No

nanonyme:
Try SHARED_LIBRARY_GUARD=0 with Steam wrapper

nightuser:
@nanonyme no, it works now. I tried specifying it on CLI before, but apparently I forgot about `env` switch and did it incorrectly.

However, if I add `--command=steam` it works regardless of what is in `SHARED_LIBRARY_GUARD`.

nanonyme:
Please do not do that. It will break your Steam deployment. Can you help me debug the actual bug?

nightuser:
@nanonyme sure.

nightuser:
@nanonyme I inspected the debug output with and without the guard, and here's the difference in logs:

With the guard:
```
Game update: AppID 589590 "", ProcID 274, IP 0.0.0.0:0
ERROR: ld.so: object '/home/nightuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
Starting app 589590
Installing breakpad exception handler for appid(steam)/version(1570046470)
>>> Adding process 274 for game ID 589590
GameAction [AppID 589590, ActionID 1] : LaunchApp changed task to WaitingGameWindow with ""
shared-library-guard active for /home/nightuser/.local/share/Steam/steamapps/common/Kindergarten/Kindergarten.x86
ERROR: ld.so: object '/home/nightuser/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
GameAction [AppID 589590, ActionID 1] : LaunchApp changed task to Completed with ""
>>> Adding process 275 for game ID 589590
[...]

Issue #484, opened 2019-10-15
EVE Online does not launch when using steam-wrapper


eskse:
**Game information**
Eve Online

**Distribution name and version where applicable**
Fedora 31 Silverblue (Beta)

**Flatpak info**
Flatpak 1.4.3
default
host

**Problem description**
When starting Steam via the desktop file or running the steam-wrapper script via the shell EVE Online loads to the point where the tray icon appears but the game launcher doesn't open. The tray icon also doesn't have any functionality when clicked.

The issue does not occur when running the steam binary directly from the flatpak shell

**Does this issue reproduce with native Steam**
Have not tested (assuming steam-wrapper is part of the flatpak)

nanonyme:
Does this now work?

eskse:
Yes it's working correctly now. Thanks!

Issue #482, opened 2019-10-14
Geometry Wars 3: Dimensions Evolved crashes while loading


Pixelnarium:
**Game information**
https://store.steampowered.com/app/310790/Geometry_Wars_3_Dimensions_Evolved/

It is a 32bit game. The game worked fine in an earlier Flatpak-Steam version (18.08 based runtime).

(inside the flatpak) $ ldd GeometryWars3
```
$ ldd GeometryWars3 
	linux-gate.so.1 (0xf7fba000)
	libX11.so.6 => /app/lib/i386-linux-gnu/libX11.so.6 (0xf7e48000)
	libGL.so.1 => /app/lib/i386-linux-gnu/libGL.so.1 (0xf7ddd000)
	libfmodex.so => ./libfmodex.so (0xf7c29000)
	libfmodevent.so => ./libfmodevent.so (0xf7b9c000)
	libpthread.so.0 => /app/lib/i386-linux-gnu/libpthread.so.0 (0xf7b79000)
	librt.so.1 => /app/lib/i386-linux-gnu/librt.so.1 (0xf7b6d000)
	libdl.so.2 => /app/lib/i386-linux-gnu/libdl.so.2 (0xf7b67000)
	libm.so.6 => /app/lib/i386-linux-gnu/libm.so.6 (0xf7a90000)
	libuuid.so.1 => /app/lib/i386-linux-gnu/libuuid.so.1 (0xf7a85000)
	libstdc++.so.6 => /app/lib/i386-linux-gnu/libstdc++.so.6 (0xf7886000)
	libc++.so.1 => ./libc++.so.1 (0xf77da000)
	libcxxrt.so => ./libcxxrt.so (0xf77bb000)
	libc.so.6 => /app/lib/i386-linux-gnu/libc.so.6 (0xf75d0000)
	libgcc_s.so.1 => /app/lib/i386-linux-gnu/libgcc_s.so.1 (0xf75b1000)
	libiconv.so.2 => ./libiconv.so.2 (0xf74cd000)
	libopenal.so.1 => /app/lib/i386-linux-gnu/libopenal.so.1 (0xf73cd000)
	libsteam_api.so => ./libsteam_api.so (0xf73bd000)
	libxcb.so.1 => /app/lib/i386-linux-gnu/libxcb.so.1 (0xf738d000)
	libGLX.so.0 => /app/lib/i386-linux-gnu/libGLX.so.0 (0xf7369000)
[...]

Issue #417, opened 2019-08-15
Games using SDL-image fail to launch


FastOSlinux:
**Game information**
Shadow of mordor

**Distribution name and version where applicable**
$ cat /etc/os-release
NAME=Fedora
VERSION="30.20190814.0 (Workstation Edition)"
ID=fedora
VERSION_ID=30
VERSION_CODENAME=""
PLATFORM_ID="platform:f30"
PRETTY_NAME="Fedora 30.20190814.0 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:30"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f30/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=30
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=30
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
OSTREE_VERSION=30.20190814.0

-->

**Flatpak info**

**Problem description**
do not open

log
 https://paste.fedoraproject.org/paste/a24qSmlATA7aZw95GkSCJw/deactivate/5zlXh97pUGG16ietkQgiTsVW3iSxjlVJhbSwCsahwp4Mccfgoy4gtqpFOsNB2A3m

**Does this issue reproduce with native Steam**

nanonyme:
I get warnings about deactivating the paste when trying to open it. Maybe use a gist instead?

FastOSlinux:
> I get warnings about deactivating the paste when trying to open it. Maybe use a gist instead?

I don't understand, but it's something with my partition, because other games in the same location are not opening

FastOSlinux:
I moved to another partition and still the same problem

Loaded Config for Local Selection Path for App ID 241930, Controller 0: /var/home/fastos/.local/share/Steam/steamapps/workshop/content/241100/485450955/701784661372096826_legacy.bin
[...]

PR #486, opened 2019-10-17
Degeneralize workaround

nanonyme:
Related to #417 #485 #484 #482


nightuser:
@nanonyme fixed my issue, thanks! How to get back to the regular stable version when the fix will be released?

nanonyme:
You can see the full ref with flatpak list. You can just uninstall the application. It will have /test at the end of the ref iirc.

PR #491, opened 2019-10-20
Update compatibility tools support

gasinvein:
Valve has added support for compatibility tools installed system-wide in current Steam Beta https://github.com/ValveSoftware/steam-for-linux/issues/6310#issuecomment-544279123
We can now stop re-writing ct manifests into users home directory via the wrapper.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -43,6 +43,7 @@ finish-args:
   - --env=SDL_VIDEODRIVER=
   - --env=DBUS_FATAL_WARNINGS=0
   - --env=SSL_CERT_DIR=/etc/ssl/certs
+  - --env=STEAM_EXTRA_COMPAT_TOOLS_PATHS=/app/share/steam/compatibilitytools.d
   - --require-version=1.0.0
 
 add-extensions:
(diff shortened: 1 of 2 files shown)


nanonyme:
Do you have some compat tool available that you can test this with?

gasinvein:
@nanonyme Yes, I'm currently testing this with Boxtron.

gasinvein:
Seems like it's working.

@nanonyme AFAIK Steam support compatibility tools in two layouts:

- one with the tools placed in a subdirectory (e.g. `compatibitytools.d/Boxtron/compatibilitytool.vdf`)
- and the other without a subdirectory (e.g.  `compatibitytools.d/boxtron.vdf`)

Should we support both? I guess Flatpak's `merge-dirs` could help with that.

nanonyme:
Dunno. If this is fine, I'm going to land this. Can be extended later. This simplifies wrapper which is always a good thing.

gasinvein:
I believe this could be merged into stable, too. "Beta" here mans Steam client beta, it's opt-in in the client regardless of the flatpak branch.

nanonyme:
I'd prefer to get that one more SDL2 quirk file added before doing another release.

PR #582, opened 2020-06-30
Migrate to 20.08

nanonyme:

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -1,6 +1,6 @@
 app-id: com.valvesoftware.Steam
 runtime: org.freedesktop.Platform
-runtime-version: '19.08'
+runtime-version: '20.08beta'
 sdk: org.freedesktop.Sdk
 command: steam-wrapper
 separate-locales: false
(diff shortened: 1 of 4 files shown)


nanonyme:
bot, build

gasinvein:
I see vkd3d fails to build. I might be a good idea to just drop it, it's not useful anymore anyway (Valve build their own incompatible fork for Proton, latest release 1.1 is just too old for anything).

valentindavid:
For vkd3d I had success with: https://github.com/valentindavid/com.valvesoftware.Steam/commit/8b9a4b793aa5fae92fbfc4880b1bf71c202a945b

I am releasing beta.2 today which will fix nss so that the browser in Steam works again.

gasinvein:
@valentindavid Now there is a hard-fork of vkd3d by Valve, specifically for Proton.

valentindavid:
@gasinvein OK, we can switch to it.

gasinvein:
@valentindavid I'd say it might be better to just drop it. Proton bundles it's own copy anyway.

valentindavid:
@gasinvein All versions of proton?

gasinvein:
@valentindavid At least since 4.11, and I'm not sure if previous versions even supported it.

nanonyme:
Okay, I'll remove it. Thanks for comments.

nanonyme:
Looking good. App will be entering bugfixes only mode once this is merged until 20.08 is out.

PR #594, opened 2020-08-01
Update FAudio to 20.08 and drop ffmpeg

gasinvein:
FAudio now uses gstreamer instead of ffmpeg.

### DIFF
--- a/modules-32bit.yml
+++ b/modules-32bit.yml
@@ -41,36 +41,13 @@ modules:
 
 # -- faudio --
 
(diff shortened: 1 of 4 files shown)


nanonyme:
Hm, so why do we need to keep shipping our replacement if it uses gstreamer anyway?

gasinvein:
Probably it's still useful for older (up to current) Proton versions. It seems like Valve doesn't update FAudio for previous releases.

PR #621, opened 2020-09-22
gperftools: update to 2.8

Lctrs:


nanonyme:
Did you check that the game is compatible with this?

Lctrs:
Nope !

Lctrs:
I don't own HL2 so I couldn't test with it, but I do own HL2: Deathmatch - which uses the same engine - and runned it without any issue with this commit.

nanonyme:
Some of these bundled gperftools, some do not.

PR #624, opened 2020-09-22
Backup config and data on migration

gasinvein:
Since we are now migrating data without asking for user consent, we should make sure that files to be overwritten aren't lost permanently. So, backup `~/.config` and `~/.local/share` if present, just in case.

Follow-up for #622

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -129,8 +129,11 @@ def migrate_config():
     """
     source = os.path.expandvars("$XDG_CONFIG_HOME")
     target = CONFIG
+    backup = f'{CONFIG}.bak'
     relocated = os.path.expandvars("$XDG_CONFIG_HOME.old")
     if not os.path.islink(source):
+        if os.path.isdir(target):
+            copytree(target, backup)
         copytree(source, target)
         os.rename(source, relocated)
         os.symlink(target, source)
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -146,9 +149,12 @@ def migrate_data():
     """
     source = os.path.expandvars("$XDG_DATA_HOME")
     target = DATA
+    backup = f'{DATA}.bak'
     steam_home = os.path.join(source, "Steam")
     xdg_data_home = os.path.join(STEAM_ROOT, target)
     if not os.path.islink(source):
+        if os.path.isdir(target):
+            copytree(target, backup)
```
gasinvein, line 157: It's not possible for the Steam directory to be in `~/.local/share/Steam` before migration, right? Otherwise, it should be `ignore=`'d here.

nanonyme: Should not be. These are paths inside sandbox

gasinvein: I can imagine 2 scenarios how it can happen:
- Steam client ignored `$XDG_DATA_HOME` and created its data dir under the fallback path `~/.local/share` (older Steam client didn't respect XDG spec?)
- `~/.var/app/com.valvesoftware.Steam/data` symlink got deleted or not created (e.g. after interrupted migration)

PR #627, opened 2020-09-29
Try to keep using atomic operations for data more

nanonyme:
Instead of making backup, let's just make sure we use enough atomic operations that if things break, we will eventually land in consistent state.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -149,18 +149,19 @@ def migrate_data():
     """
     source = os.path.expandvars("$XDG_DATA_HOME")
     target = DATA
-    backup = f'{DATA}.bak'
-    steam_home = os.path.join(source, "Steam")
+    staging = f'{DATA}.staging'
     xdg_data_home = os.path.join(STEAM_ROOT, target)
     if not os.path.islink(source):
-        if os.path.isdir(target):
-            copytree(target, backup)
-        copytree(source, target, ignore=[steam_home])
+        if os.path.exists(source):
+            os.rename(source, staging)
+        os.symlink(target, source)
+    if os.path.isdir(staging):
+        steam_home = os.path.join(staging, "Steam")
+        copytree(staging, target, ignore=[steam_home])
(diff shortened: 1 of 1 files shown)


nanonyme:
@gasinvein do you have time to review? I rewrote the entire data migration part to try to make more parts of it atomic.

gasinvein:
Can you please explain when and why does copying fail currently?

nanonyme:
@gasinvein the copytree fails on Steam Runtime iirc (the Ubuntu stuff inside Steam installation directory).

gasinvein:
You mean during backup stage (i.e. since c25fdcadcb2dfeac6f297c2cbc47f28b8390dfbb), or else?
It didn't fail for me neither before nor after recent changes. Yet I have no games in the default library, not sure if that matters.

nanonyme:
At least those were the experiences way back when this was first written. Anyway, the intent is we overwrite target with what was in source. Target can exist, that is fine. We don't need backup at that point. The problem is the source which needs to be replaced with a symlink. If you start Steam without steam-wrapper, then Steam will write there.

nanonyme:
I'm actually potentally also confused. Maybe target is smaller and safe to just backup.

nanonyme:
I'll do one round of testing. If everything seems good, I'll roll a release still today without these changes.

nanonyme:
Turns out the migration doesn't actually work as is. But I don't think your change broke it any worse.

gasinvein:
Sorry, I fail to understand how this PR changes target-related behavior. In both cases we end up `cypytree`'ing `~/.var/app/com.valvesoftware.Steam/data` to `~/.var/app/com.valvesoftware.Steam/.local/share`, except for that now we move the former aside before copying.

nanonyme:
Basically the idea in this PR was doing the symlink *first*. If the older code fails at making the symlink (which should not break hopefully?), it will already have migrated everything so next run will copy dozens of gigabytes of files to a new directory.

nanonyme:
Closing, will proceed with a release.

PR #628, opened 2020-09-29
Release beta to stable

nanonyme:
Some updates and cleanup of cloud migration code

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -129,8 +129,11 @@ def migrate_config():
     """
     source = os.path.expandvars("$XDG_CONFIG_HOME")
     target = CONFIG
+    backup = f'{CONFIG}.bak'
     relocated = os.path.expandvars("$XDG_CONFIG_HOME.old")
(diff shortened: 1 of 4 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -181,9 +149,12 @@ def migrate_data():
     """
     source = os.path.expandvars("$XDG_DATA_HOME")
     target = DATA
+    backup = f'{DATA}.bak'
     steam_home = os.path.join(source, "Steam")
     xdg_data_home = os.path.join(STEAM_ROOT, target)
     if not os.path.islink(source):
+        if os.path.isdir(target):
+            copytree(target, backup)
```
gasinvein, line 157: We should either exclude `.local/share/Steam` here or do not backup data at all. Otherwise, we can end up backing up the whole Steam library.

nanonyme: There's now code handling this scenario.

nanonyme:
Tested, seems fine. Merging.

PR #629, opened 2020-09-29
Exclude Steam directory itself from backup

nanonyme:

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -154,7 +154,8 @@ def migrate_data():
     xdg_data_home = os.path.join(STEAM_ROOT, target)
     if not os.path.islink(source):
         if os.path.isdir(target):
-            copytree(target, backup)
+            copytree(target, backup,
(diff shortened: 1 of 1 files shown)


nanonyme:
@gasinvein created this

steam_wrapper/steam_wrapper.py
```diff
@@ -154,7 +154,8 @@ def migrate_data():
     xdg_data_home = os.path.join(STEAM_ROOT, target)
     if not os.path.islink(source):
         if os.path.isdir(target):
-            copytree(target, backup)
+            copytree(target, backup,
+                     ignore=[os.path.join(target, "Steam")])
```
gasinvein, line 158: `target` here is relative path. Will `fnmatch()` handle it correctly?

nanonyme: It should. fnmatch won't care even if the file doesn't exist.

gasinvein: Looks fine to me, then.

PR #632, opened 2020-10-05
Bump Python version for util extensions

gasinvein:
I guess this was overlooked during SDK update: python version changed from 3.7 to 3.8.


nanonyme:
Right, it's a bit annoying this is hardcoded. Oh, well.

gasinvein:
Hm. Why we even use merge-dirs for that? Extending `PYTHONPATH` with proper per-extension values should be trivial in python-based wrapper.

PR #648, opened 2020-11-08
Fix str.lstrip() misuse in wrapper

gasinvein:
Remove extension point prefix from extension ID properly.
In #633, I've misused Python's `str.lstrip()` method to remove leading character sequence, which is not suitable for that.
This could result in wrong (cropped) extension directory names, which would result in extensions not actually enabled. Luckily this didn't fire yet.


gasinvein:
Closing in favor of #652

PR #649, opened 2020-11-09
Release beta to stable

nanonyme:


gasinvein:
Extension version should be changed to `stable`, otherwise `stable` extensions won't work.
This might require additional commit in git master only.

Issue #559, opened 2020-05-06
Add PrefersNonDefaultGPU=true to .desktop file of Steam


Eonfge:
Duplicate of the upstream issue:
https://github.com/ValveSoftware/steam-for-linux/issues/7089

---

There is a new standard for aiding in multi-GPU setups. You can now set a flag called `PrefersNonDefaultGPU` which will tell the computer to use the alternative (generally speaking, dedicated) GPU. 

By setting this flag, Steam and it's applications will run with the more powerful GPU by default, preventing many annoyed customers as they are rushing B with 20 fps.

#### Docs
- https://www.gamingonlinux.com/articles/the-linux-desktop-entry-specification-gets-a-way-to-automatically-use-a-discrete-gpu-merged-into-gnome.16598
- https://www.hadess.net/2020/05/dual-gpu-support-launch-on-discrete-gpu.html
- https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

---

If upstream is not adding the feature, I would propose you guys add it manually.

nanonyme:
Have you tested that it actually makes the games Steam launch run in the other GPU, not just Steam itself? Ideally Steam would run on primary GPU and games on secondary more powerful one.

Eonfge:
Can confirm. 

What I do:
- Search for Flatpak Steam
- Right click icon and select 'launch using Dedicated Graphics Card'
- Test a game, in this case Mindustry

What I see using Green With Envy, is that steam itself only takes a bit of video memory. Then when I start Mindustry, video game memory usage jumps up and the card starts actively running.

crawfxrd:
```
Tue May 12 14:55:09 2020       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 440.82       Driver Version: 440.82       CUDA Version: 10.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce RTX 207...  Off  | 00000000:01:00.0 Off |                  N/A |
| N/A   44C    P8     8W /  N/A |    186MiB /  7982MiB |     38%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
[...]

PR #653, opened 2020-11-12
Install prime-run helper

ranisalt:
This script runs a command with the dedicated GPU on hybrid NVIDIA systems. It is GPL-licensed and comes from https://github.com/archlinux/svntogit-packages/tree/packages/nvidia-prime/trunk

I know #559 already helps but even with that, my dedicated GPU will still not be used in games, it only works with the helper scripts which set a couple of variables. I could copy and paste these to all games but `prime-run %command%` is much cleaner.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -219,3 +219,11 @@ modules:
     sources:
       - type: dir
         path: resources
+
+  - name: prime-run
+    buildsystem: simple
+    sources:
+      - type: file
+        path: prime-run
+    build-commands:
+      - install -Dm755 -t /app/bin prime-run
--- a/prime-run
+++ b/prime-run
@@ -1,0 +1,2 @@
+#!/bin/bash
+__NV_PRIME_RENDER_OFFLOAD=1 __VK_LAYER_NV_optimus=NVIDIA_only __GLX_VENDOR_LIBRARY_NAME=nvidia "$@"


nanonyme:
I would prefer using that repo as source instead of bundling so it's more obvious where it comes from.

nanonyme:
Looks like https://github.com/amonakov/primus is the upstream

ranisalt:
@nanonyme well, that is just a package from Arch Linux and nothing official or such. But I can do it

gasinvein:
Honestly I don't think we need a dedicated wrapper for this. It's way easier to just set the environment variables as flatpak overrides.

gasinvein:
Also since the wrapper just sets variables, it should also work outside flatpak sandbox, i.e. `prime-run flatpak com.valvesoftware.Steam` should work.

ranisalt:
> Also since the wrapper just sets variables, it should also work outside flatpak sandbox, i.e. `prime-run flatpak run com.valvesoftware.Steam` should work.

Yes, but that will run the Steam client itself in the dedicated GPU, which will turn it on even if I just want to browse the store

nanonyme:
License-wise I think this is not really ok. Arch Linux is not using GPL correctly as it requires recipient to get a copy of the license along with the program. And the derivative in this repo strips even the lighter mention (which is legally insufficient) that was in the PKGBUILD file that this would be GPL.

PR #655, opened 2020-11-16
Build libusb both 32/64bit and keep devel files

gasinvein:
It is used by Wine, so, we need both 32 and 64-bit libusb, and also keep development files to build Proton.

libusb isn't updated frequently (latest version was released a year ago) , so I believe it won't add much maintenance burden.

### DIFF
--- a/modules-32bit.yml
+++ b/modules-32bit.yml
@@ -27,6 +27,8 @@ modules:
     build-options:
       arch:
         x86_64: *compat_i386_opts
+      ldflags: -L/app/lib32 -Wl,-z,relro,-z,now -Wl,--as-needed
+      ldflags-override: true
     sources:
       - sources/libusb-archive.json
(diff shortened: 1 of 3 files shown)


nanonyme:
Not entirely happy since this introduces ABI contract between app and its extensions.

gasinvein:
Well, we can bundle libusb in Proton, but since it's needed in Steam bundle as well - I thought it should be here.

nanonyme:
Sure. I'm still not happy about the ABI dependency. Those tend to bite when you least expect it.

gasinvein:
Hm, what if we copy libusb to Proton bundle at build time? I mean, `cp /app/lib/libusb* $FLATPAK_DEST/lib/` I guess the ABI will remain then in case if libusb changes in Steam bundle.

PR #659, opened 2020-11-21
Add ffmpeg-full extensions

gasinvein:
#594 removed bundled ffmpeg, since FAudio doesn't use it directly anymore, but it appears that we still need codecs, for gst-libav to be fully functional.
Together with flathub/com.valvesoftware.Steam.CompatibilityTool.Proton#21 should fix sound in some xaudio-based games.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -149,7 +149,6 @@ modules:
         path: steam_wrapper
     build-commands:
       - python3 -mpip install . --prefix=/app --no-index --find-links .
-    modules:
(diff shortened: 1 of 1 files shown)


nanonyme:
Oops. Forgot to check if autodownload is off. Can you disable? There should be user consent for installing ffmpeg-full

gasinvein:
Are you it shouldn't be auto-downloaded? There is no easy way to tell user to install this extensions, and gstreamer error messages are rather obscure in case of missing codecs in ffmpeg.

PR #663, opened 2020-11-25
Add 32-bit gstreamer plugins paths

gasinvein:
Default `GST_PLUGIN_SYSTEM_PATH` includes only native plugins.
This change makes sure that all present codecs are available to 32-bit games as well.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -57,6 +57,7 @@ finish-args:
   - --env=PYTHONPATH=/app/utils/lib/python3.8/site-packages
   - --env=PROTON_DEBUG_DIR=/var/tmp
   - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
   - --require-version=1.0.0
 
 add-extensions:


nanonyme:
Are you sure this is needed? See https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/release/20.08/patches/gstreamer-libav/gst-libav-stop-caching-codecs.patch

gasinvein:
This is about gstreamer plugins, not ffmpeg. With `GST_PLUGIN_SYSTEM_PATH` set to its current default value, 32-bit `gst-inspect-1.0` doesn't list any modules except `bin` and `pipeline`.

nanonyme:
Do you have any to test with? I'm mostly worried about regressions here. This environment variable is used by a lot of things.

gasinvein:
Things that used to work before seem to keep working for me, namely Skyrim SE.

com.valvesoftware.Steam.yml
```diff
@@ -57,6 +57,7 @@ finish-args:
   - --env=PYTHONPATH=/app/utils/lib/python3.8/site-packages
   - --env=PROTON_DEBUG_DIR=/var/tmp
   - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
```
gasinvein, line 60: Maybe we should remove `/app/lib{,32}/gstreamer-1.0` from the list, since we don't bundle any gstreamer plugins?

nanonyme: I'm fine with it as is. We can leave it as is so it will work sanely if we ever bundle any plugins.

PR #678, opened 2020-12-25
Allow UDisks2 system talk name

gasinvein:
Wine uses UDisks to enumerate drives (backing `mountmgr.sys` implementation). This is mostly used in various installers and unlikely to be needed for games at run time, but since it can break things with very obscure errors (wine itself doesn't complain when it can't talk to udisks), I believe it's better to have UDisks2 available to Wine.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -24,6 +24,8 @@ finish-args:
   # connecting to wifi and so on.
   # In normal mode, it probably just gets info about current connection status.
   - --system-talk-name=org.freedesktop.NetworkManager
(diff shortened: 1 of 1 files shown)


nanonyme:
What's the impact with regards to hole in sandbox? What can you do through this?

gasinvein:
It depends on polkit setting on host. In general, I believe UDisks2 doesn't allow anything unsafe/insecure without user authorization (with password prompt). It may allow mounting removable devices and something like that without authorization, though.

nanonyme:
Hm, right. Those will not be visible within the sandbox anyhow.

PR #684, opened 2020-12-30
Allow user to override xdg paths

### LINKED ISSUE #683: Bad item [...] found in filesystem overrides
After the latest update I can no longer launch Steam due to the following error:

```
Bad item "xdg-config/gtk-3.0:ro" found in filesystem overrides
```

Yes, I do have such an override, and it's made on purpose for theming-related stuff.
I'm curious, why does the wrapper care about that, and why does it prevent Steam from starting?

nanonyme:
Fixes #683

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -95,7 +95,7 @@ def check_bad_filesystem_entries(entries):
                  "host",
                  os.path.expandvars("/var/home/$USER"),
                  os.path.expandvars("/home/$USER")]
-    bad_topdirs = ["xdg-config", "xdg-data", "xdg-cache"]
+    bad_topdirs = ["home/", "system/"]
     found = False
     for entry in entries:
         assert ";" not in entry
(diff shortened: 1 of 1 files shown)


nanonyme:
Ping @gasinvein can you check this out? It seems to work correctly for me.

gasinvein:
Before going into implementation details, I'd like to clarify the approach in general. If we just skip config/data/cache migration when there are fs overrides under it, migration won't be ever applied for users who added such overrides before first launch.
E.g. if user had `--filesystem=xdg-config/gtk-3.0` global override before installing Steam, `$XDG_CONFIG_HOME` and `~/.config` will stay as separate dirs for them until the override is removed. I'm not sure if this is what we want?

nanonyme:
@gasinvein there is no separate dir. If you have the filesystem override, $XDG_CONFIG_HOME will point to ~/.config.

nanonyme:
This is exactly why the migraions have been failing thus far. We were telling we want migration to be done from ~/.config to ~/.config. That will not end well for obvious reasons.

nanonyme:
Okay, so the assumption was wrong. The paths are very much still original. I'm also really now suddenly a bit worried what happens if xdg-config path in host is config. We definitely should not migrate in that case.

PR #691, opened 2021-01-05
Allow selectively passing through xdg-config

nanonyme:
User should be able to expose select parts of xdg-config-home to app. When exposed, we must not overwrite those. So calculate which items are exposed and skip processing them. When user next time starts app, filesystems get mounted properly on top of directory structure anyway.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -6,6 +6,7 @@ import shutil
 import errno
 import fnmatch
 import configparser
+import re
 from distutils.version import LooseVersion
 
 
@@ -96,7 +97,7 @@ def check_bad_filesystem_entries(entries):
                  "host",
                  os.path.expandvars("/var/home/$USER"),
                  os.path.expandvars("/home/$USER")]
-    bad_topdirs = ["xdg-config", "xdg-data", "xdg-cache"]
+    bad_topdirs = ["xdg-data", "xdg-cache"]
     found = False
     for entry in entries:
         assert ";" not in entry
@@ -125,7 +126,18 @@ def check_allowed_to_run(current_info):
         shutil.rmtree(steam_home)
 
 
-def migrate_config():
+def make_xdg_ignores(current_info, xdg_name, path):
+    ignores = []
+    filesystems = current_info["filesystems"]
+    for filesystem in filesystems:
+        filesystem_path = filesystem.split(":")[0]
+        if filesystem_path.startswith(xdg_name):
+            ignores.append(re.sub(f"^{xdg_name}", path,
+                                  filesystem_path))
+    return ignores
+
+
+def migrate_config(current_info):
     """
     There's bind-mounted contents inside config dir so we need to
     1) Relocate, move to temp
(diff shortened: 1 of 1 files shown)


nanonyme:
@gasinvein so here's another iteration of this

gasinvein:
Can you please explain high-level goal here? You want to exclude subpaths added via overrides from migration, right?

nanonyme:
@gasinvein is the description now good?

steam_wrapper/steam_wrapper.py
```diff
@@ -125,7 +126,18 @@ def check_allowed_to_run(current_info):
...
 
 
-def migrate_config():
+def make_xdg_ignores(current_info, xdg_name, path):
+    ignores = []
+    filesystems = current_info["filesystems"]
+    for filesystem in filesystems:
+        filesystem_path = filesystem.split(":")[0]
+        if filesystem_path.startswith(xdg_name):
+            ignores.append(re.sub(f"^{xdg_name}", path,
```
gasinvein, line 135: Why use regex here? Maybe just use `os.path.split()` here to turn string into sequence of path components?

steam_wrapper/steam_wrapper.py
```diff
@@ -96,7 +97,7 @@ def check_bad_filesystem_entries(entries):
                  "host",
                  os.path.expandvars("/var/home/$USER"),
                  os.path.expandvars("/home/$USER")]
-    bad_topdirs = ["xdg-config", "xdg-data", "xdg-cache"]
+    bad_topdirs = ["xdg-data", "xdg-cache", "home", "host"]
```
gasinvein, line 100: AFAIK `home` and `host` are special values that can't be used in paths (e.g. `--filesystem=home/.config` won't work). Needs confirmation.

nanonyme: They can, this was changed in a recent PR.

nanonyme: Ah, okay, I thought it was both.

steam_wrapper/steam_wrapper.py
```diff
@@ -125,7 +126,18 @@ def check_allowed_to_run(current_info):
...
 
-def migrate_config():
+def make_xdg_ignores(current_info, xdg_name, path):
+    ignores = []
+    filesystems = current_info["filesystems"]
+    for filesystem in filesystems:
+        filesystem_path = filesystem.split(":")[0]
+        if filesystem_path.startswith(xdg_name):
+            ignores.append(re.sub(f"^{xdg_name}", path,
+                                  filesystem_path))
```
gasinvein, line 136: I've meant something like this when suggested `os.path.split()` above (didn't actually test it).
```suggestion
        filesystem_path = filesystem.rsplit(':', 1)[0]
        path_seq = os.path.split(filesystem_path)
        if path_seq[0] == xdg_name:
            ignores.append(os.path.join(*path_seq[1:]))
```

nanonyme:
Closing in favour of https://github.com/flathub/com.valvesoftware.Steam/pull/698/files#diff-1ef8118377e63a013152accdfca3de947521d604001dc2a5271b20e41448db6bL95

PR #672, opened 2020-12-13
Add more filesystem checks to wrapper

gasinvein:
This should prevent same-file error on copying xdg-dirs during migration.
Steam will refuse to launch for users who added overrides like https://github.com/flathub/com.valvesoftware.Steam/issues/667#issuecomment-743480550. This is intentional.
Closes #667

PR #698, opened 2021-01-09
Migrate absolute paths under flatpak state dir

gasinvein:
The idea is not to mess with files outside the flatpak state directory, e.g. migrate `~/.var/app/com.valvesoftware.Steam/config` to `~/.var/app/com.valvesoftware.Steam/.config` instead of `~/.config`. This way we won't cross mount point boundaries under destination directory.
Makes #672 and #691 obsolete.

This is ready for review, but **requires excessive testing**.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -96,12 +96,11 @@ def check_bad_filesystem_entries(entries):
                  "host",
                  os.path.expandvars("/var/home/$USER"),
                  os.path.expandvars("/home/$USER")]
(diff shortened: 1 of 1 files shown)


nanonyme:
I think this looks more or less fine. Basically way to test this is
`flatpak run --command=bash com.valvesoftware.Steam`
then run `steam` (not wrapper) and install some games. Then launch app normally and see files are correctly migrated.

gasinvein:
Yeah, I've tested the simplest case with running `steam` without wrapper first, then applying migration.

PR #701, opened 2021-01-12
Migrate xdg-data in two stages

gasinvein:
Since there could be mount points under `xdg-data`, we probably should move the data directory aside on first launch, and delete it on second, similarly to how we handle `xdg-config` currently.
Until we properly exclude xdg-dirs passed from host via overrides, this should make the wrapper a bit safer.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -162,6 +162,7 @@ def migrate_data():
     source = os.path.expandvars("$XDG_DATA_HOME")
     target = os.path.join(FLATPAK_STATE_DIR, DEFAULT_DATA_DIR)
     backup = f'{target}.bak'
+    relocated = os.path.expandvars("$XDG_DATA_HOME.old")
     steam_root_source = os.path.join(source, "Steam")
     steam_root_target = os.path.join(target, "Steam")
     if not os.path.islink(source):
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -170,8 +171,11 @@ def migrate_data():
         copytree(source, target, ignore=[steam_root_source])
         if os.path.isdir(steam_root_source):
             os.rename(steam_root_source, steam_root_target)
```
nanonyme, line 173: Won't this still fail with overrides? I actually wonder if we could do migrate, then tell flatpak to restart the app at the end.

gasinvein: It will fail on `xdg-data/Steam` override, yes.

gasinvein: Actually I wonder why we can rename the parent directory of the mount point, but not the mount point itself. Anyway, I still consider passing `~/.local/share/Steam` from host a bad idea (it is not shareable with non-flatpak steam).

nanonyme: Then IMHO we need to check and block. The failure is quite confusing. Also I don't like the idea of *copying* mounted things. I think I did it wrong before as well. I think ideally

1. Copy things where not a mount
2. Move directory aside
3. Create symlink
4. execv flatpak-spawn the same script so we get a new sandbox
5. Remove the directory that was moved aside
6. Continue

gasinvein: Err, this PR wasn't about handling the steam root dir, it was for safely handling cases like `xdg-data/icons`.
I agree that we need to exclude mounted dirs, it's just out of scope here.

nanonyme: Ok.

nanonyme: I'm exposing the restart in https://github.com/flathub/com.valvesoftware.Steam/pull/702; can you rebase this after I merge that and hook into the restart mechanism?

PR #702, opened 2021-01-13
Restart app if we just migrated config

nanonyme:
Make sure we have consistent sandbox on first startup of Steam

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -196,13 +196,17 @@ class Migrator:
         assert self.need_cleanup
         self.log.info(f"Deleting {self.relocated_source}")
         shutil.rmtree(self.relocated_source)
-        return [self.relocated_source]
 
     def apply(self):
+        """
+        Return value means whether we need app restart
+        """
         if self.need_migration:
             self.do_migrate()
+            return self.two_steps
         elif self.need_cleanup:
             self.do_cleanup()
+        return False
 
 
 def migrate_config():
@@ -215,8 +219,9 @@ def migrate_config():
     migrator = Migrator(os.path.expandvars("$XDG_CONFIG_HOME"),
                         os.path.join(FLATPAK_STATE_DIR, DEFAULT_CONFIG_DIR),
                         two_steps=True)
-    migrator.apply()
+    should_restart = migrator.apply()
     os.environ["XDG_CONFIG_HOME"] = os.path.expandvars(f"$HOME/{DEFAULT_CONFIG_DIR}")
+    return should_restart
 
 
 def migrate_data():
@@ -228,16 +233,18 @@ def migrate_data():
                         os.path.join(FLATPAK_STATE_DIR, DEFAULT_DATA_DIR),
                         two_steps=True,
                         rename=["Steam"])
-    migrator.apply()
+    should_restart = migrator.apply()
     os.environ["XDG_DATA_HOME"] = os.path.expandvars(f"$HOME/{DEFAULT_DATA_DIR}")
+    return should_restart
 
 
 def migrate_cache():
     migrator = Migrator(os.path.expandvars("$XDG_CACHE_HOME"),
                         os.path.join(FLATPAK_STATE_DIR, DEFAULT_CACHE_DIR),
                         need_backup=False)
-    migrator.apply()
+    should_restart = migrator.apply()
     os.environ["XDG_CACHE_HOME"] = os.path.expandvars(f"$HOME/{DEFAULT_CACHE_DIR}")
+    return should_restart
 
 
 def enable_discord_rpc():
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -215,8 +216,10 @@ def migrate_config():
     migrator = Migrator(os.path.expandvars("$XDG_CONFIG_HOME"),
                         os.path.join(FLATPAK_STATE_DIR, DEFAULT_CONFIG_DIR),
                         two_steps=True)
-    migrator.apply()
-    os.environ["XDG_CONFIG_HOME"] = os.path.expandvars(f"$HOME/{DEFAULT_CONFIG_DIR}")
+    should_restart = migrator.apply()
+    if not should_restart:
```
gasinvein, line 220: I'm not sure why we need this check? AFAIK `flatpak-spawn` resets environment anyway.

nanonyme: Because we want to avoid infinite sandbox recursion. We need to stop re-launching the wrapper script in new sandbox when migration is done.

gasinvein: But setting `XDG_CONFIG_HOME` doesn't affect restart.

steam_wrapper/steam_wrapper.py
```diff
@@ -196,13 +197,13 @@ def do_cleanup(self):
         assert self.need_cleanup
         self.log.info(f"Deleting {self.relocated_source}")
         shutil.rmtree(self.relocated_source)
-        return [self.relocated_source]
+        return False
```
gasinvein, line 200: I'd rather avoid static return values. Just don't check it in `apply()` if it doesn't vary.

nanonyme: I'm treating it as integer, every branch must return boolean.

nanonyme: The return value seemed not to be used anywhere. Now the return value is considered "should_restart"  boolean everywhere.

gasinvein: I got that. Just suggesting to to return `self.two_steps` value from `apply()` method after migration is applied, and `False` otherwise, and not change other methods. IMO it would be easier to understand what's going on this way.
Like this:
```python
    def apply(self):
        if self.need_migration:
            self.do_migrate()
            return self.two_steps
        elif self.need_cleanup:
            self.do_cleanup()
        return False
```

steam_wrapper/steam_wrapper.py
gasinvein: Probably my comment should've been attached to the next line. Actual question if: why set `XDG_CONFIG_HOME` only if we don't need restart?

steam_wrapper/steam_wrapper.py
```diff
@@ -196,13 +196,17 @@ def do_cleanup(self):
...
         shutil.rmtree(self.relocated_source)
-        return [self.relocated_source]
 
     def apply(self):
+        """
+        Return value means whether we need app restart
+        """
         if self.need_migration:
             self.do_migrate()
+            return self.two_step
```
gasinvein, line 206: A typo? Should be two_step**s**

steam_wrapper/steam_wrapper.py
nanonyme: Tested, not needed.

steam_wrapper/steam_wrapper.py
```diff
@@ -266,10 +272,14 @@ def main(steam_binary=STEAM_PATH):
     logging.info("https://github.com/flathub/com.valvesoftware.Steam/wiki")
     current_info = read_flatpak_info(FLATPAK_INFO)
     check_allowed_to_run(current_info)
-    migrate_config()
-    migrate_data()
+    should_restart = migrate_config()
+    should_restart += migrate_data()
     migrate_cache()
```
gasinvein, line 277: I thing we should do the same for cache just for symmetry. Also migrating cache in two steps, too, may be a good idea, but this can be considered later.

steam_wrapper/steam_wrapper.py
```diff
@@ -266,10 +272,14 @@ def main(steam_binary=STEAM_PATH):
...
-    timezone_workaround()
-    configure_shared_library_guard()
-    enable_discord_rpc()
-    os.execv(steam_binary, [steam_binary] + sys.argv[1:])
+    if should_restart:
+        command = ["/usr/bin/flatpak-spawn"] + sys.argv
```
gasinvein, line 279: Better log that we're going to restart the app.

nanonyme: Added logging

steam_wrapper/steam_wrapper.py
nanonyme: Yes. Fixed.

steam_wrapper/steam_wrapper.py
nanonyme: Fixed.

steam_wrapper/steam_wrapper.py
```diff
@@ -266,10 +273,15 @@ def main(steam_binary=STEAM_PATH):
...
-    timezone_workaround()
-    configure_shared_library_guard()
-    enable_discord_rpc()
-    os.execv(steam_binary, [steam_binary] + sys.argv[1:])
+    should_restart = migrate_config()
+    should_restart += migrate_data()
+    should_restart += migrate_cache()
+    if should_restart:
+        command = ["/usr/bin/flatpak-spawn"] + sys.argv
+        logging.info("Restarting app due to finalize sandbox tuning")
```
gasinvein, line 281: I guess the log should go _before_ the actual restart.

PR #703, opened 2021-01-13
Unify migration logic

gasinvein:
Essentially this just merges three distinct functions in one method.
Additionally, it replaces `print()`s with logging, spitting out more messages about what this script is going to do while not printing a message on each copied file.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -7,6 +7,8 @@ import errno
 import fnmatch
 import configparser
 from distutils.version import LooseVersion
+import typing as t
+import logging
 
 
 STEAM_PATH = "/app/bin/steam"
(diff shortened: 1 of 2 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -211,7 +262,8 @@ def configure_shared_library_guard():
 
 def main(steam_binary=STEAM_PATH):
     os.chdir(os.environ["HOME"]) # Ensure sane cwd
-    print ("https://github.com/flathub/com.valvesoftware.Steam/wiki")
+    logging.basicConfig(level=logging.DEBUG)
```
nanonyme, line 265: Why DEBUG by default?

gasinvein: We don't (and can't) have CLI options to increase verbosity, to I thought we should log everything by default.

nanonyme:
Overall this looks fine and I think we can add the restarting logic easily on top of this.

steam_wrapper/steam_wrapper.py
nanonyme: Well, we could have an environment variable.

nanonyme: I suppose this can be tweaked later.

PR #704, opened 2021-01-13
Skip mounts

nanonyme:
If any xdg path overrides are added, put them into ignore.

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -202,29 +202,45 @@ class Migrator:
         return False
 
 
-def migrate_config():
+
+def _get_host_xdg_mounts(xdg_name: str, flatpak_info):
+    assert xdg_name in ["xdg-data", "xdg-config", "xdg-cache"]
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -209,29 +202,45 @@ def apply(self):
...
 
 
-def migrate_config():
+
+def _get_host_xdg_mounts(xdg_name: str, flatpak_info):
+    assert xdg_name in ["xdg-data", "xdg-config", "xdg-cache"]
+    dirs: t.Set[str] = set()
+    for filesystem in flatpak_info["filesystems"]:
+        filesystem_path = filesystem.rsplit(":", 1)[0]
+        path_seq = os.path.split(filesystem_path)
```
gasinvein, line 211: `os.path.split()` is misused here. Will work for first-level subdirs, break otherwise. Sorry for bad suggestion. Here should be something like
```suggestion
        path_seq = os.path.normpath(filesystem_path).split(os.path.sep)
```

PR #705, opened 2021-01-14
Fix up skipping mounts

gasinvein:
Follow-up for #704

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -202,13 +202,12 @@ class Migrator:
         return False
 
 
-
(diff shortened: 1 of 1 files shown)


nanonyme:
What kind of a case did this break?

gasinvein:
2nd and further level subdirs weren't detected as mounts, e.g. `xdg-data/icons/hicolor` never matched `xdg-data` prefix.

nanonyme:
Sorry but I don't get how normpath helps there.

gasinvein:
It doesn't, it's for sanitizing the path just in case. Important thing is `filesystem_path.split('/')`, i.e. we split a string by slash.

PR #710, opened 2021-01-30
Make xdg dirs path prefix changeable

### LINKED ISSUE #480: "Browse local files" button in game properties in Steam client does not opens the file manager
**Game information**

This is Steam client problem. Pressing "Browse local files" button does nothing.
It is similar to https://github.com/flathub/com.valvesoftware.Steam/issues/73 but the dbus-monitor output is slightly different

```
method call time=1570995096.016638 sender=:1.4793 -> destination=org.freedesktop.DBus serial=14 path=/; interface=org.freedesktop.DBus; member=GetNameOwner
   string "org.freedesktop.DBus"
```
It does not mentions Steam.

```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

**Flatpak info**

Flatpak 1.5.0

nvidia-435-21
default
host

Have not tested, should not be an issue.

gasinvein:
Resolves #480 to some extent.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -51,6 +51,7 @@ finish-args:
   - --env=MESA_GLSL_CACHE_DIR=
   - --env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES=
   - --env=STEAM_RUNTIME=
+  - --env=FLATPAK_STEAM_XDG_DIRS_PREFIX=~/.var/app/com.valvesoftware.Steam
   - --env=SDL_VIDEODRIVER=
   - --env=DBUS_FATAL_WARNINGS=0
   - --env=SSL_CERT_DIR=/etc/ssl/certs
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -129,7 +129,7 @@ class Migrator:
     def __init__(self, source: str, target: str,
                  ignore: t.Optional[t.Set[str]]=None,
                  rename: t.Optional[t.List[str]]=None,
-                 two_steps=False, need_backup=True):
+                 two_steps=True, need_backup=True):
         self.source = source
         assert os.path.isabs(self.source)
         self.target = target
@@ -223,8 +223,7 @@ def migrate_config(flatpak_info):
     ignore = _get_host_xdg_mounts("xdg-config", flatpak_info)
     migrator = Migrator(os.path.expandvars("$XDG_CONFIG_HOME"),
                         os.path.join(FLATPAK_STATE_DIR, DEFAULT_CONFIG_DIR),
-                        ignore=ignore,
-                        two_steps=True)
+                        ignore=ignore)
     should_restart = migrator.apply()
     os.environ["XDG_CONFIG_HOME"] = os.path.expandvars(f"$HOME/{DEFAULT_CONFIG_DIR}")
     return should_restart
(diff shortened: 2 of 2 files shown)


gasinvein:
Here a my findings:
On first launch with xdg dirs shifted to `~/.var/com.valvesoftware.Steam/.local/share` etc, Steam didn't seem to care at all and didn't change symlinks in `~/.steam`. Then I deleted the `~/.steam` dir (thus forcing Steam to re-create it) and launched it again - then Steam was able to properly open file managed at screenshots dir (`~/.var/app/com.valvesoftware.Steam/.local/share/Steam/userdata/...`).

nanonyme:
Okay. I think it's probably fine to tell user to delete the symlinks. It might be too magical to try to figure out if they point to wrong places and remove it ourselves. Or?

gasinvein:
>It might be too magical to try to figure out if they point to wrong places and remove it ourselves.

This should be rather easy to implement - just read each symlink in `~/.steam` and check if it points to a dir under the xdg path prefix. If it's not - re-create it with proper target.
I'm not sure, though, if it's entirely safe to do this.

nanonyme:
I don't think we should touch the symlinks other than that if we detect .steam is stale, remove entirr directory and let Steam recreate it. But let's maybe not do it in first iteration.

nanonyme:
Otherwise this looks good to me.

gasinvein:
Removing `~/.steam` resets some setting and forces new login.

nanonyme:
Okay, let's not do that then.

gasinvein:
So how do we go then? Only change the env vars for now?

nanonyme:
Yes, I think this is good. We can suggest user to try removing .steam directory if they need this Right Now.

gasinvein:
I have more or less working implementation of "symlink shifting" under `~/.steam`, but it would require much more careful testing. I'll submit it as a follow-up PR.

com.valvesoftware.Steam.yml
```diff
@@ -51,6 +51,7 @@ finish-args:
   - --env=MESA_GLSL_CACHE_DIR=
   - --env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES=
   - --env=STEAM_RUNTIME=
+  - --env=STEAM_FLATPAK_XDG_DIRS_PREFIX=~/.var/app/com.valvesoftware.Steam
```
gasinvein, line 54: Probably we shouldn't occupy `STEAM_` prefix?

nanonyme: Might as well just XDG_DIRS_PREFIX. This takes precedence over user environment variables.

nanonyme: But this is fine. I don't think it's harmful to just keep as is.

gasinvein: I believe Steam can have special treatment of it's own variables. Maybe `FLATPAK_STEAM_XDG_DIRS_PREFIX`?

gasinvein: Changed to `FLATPAK_STEAM_XDG_DIRS_PREFIX` just in case.

PR #713, opened 2021-02-14
eudev: update to 3.2.10 and switch to official tarball

Lctrs:


nanonyme:
Is there a benefit in not using Github? I would trust its availability more in this case.

Lctrs:
@nanonyme I just tend to prefer official source, but I can go back to Github if you prefer.

PR #715, opened 2021-02-14
libbsd: update to 0.11.3

Lctrs:


nanonyme:
Does this have same soname?

Lctrs:
Yes. Before `libbsd.so.0.10.0` and after `libbsd.so.0.11.3`.

Issue #322, opened 2019-02-23
Steam tray icon is empty when using (K)StatusNotifierItem/AppIndicator Support extension


AsciiWolf:
**Distribution name and version where applicable**
NAME=Fedora
VERSION="29.20190223.0 (Workstation Edition)"
ID=fedora
VERSION_ID=29
VERSION_CODENAME=""
PLATFORM_ID="platform:f29"
PRETTY_NAME="Fedora 29.20190223.0 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:29"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f29/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=29
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=29
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
OSTREE_VERSION=29.20190223.0

**Flatpak info**
Flatpak 1.2.3

**Problem description**
The Steam tray icon seems to be empty when using the [(K)StatusNotifierItem/AppIndicator Support](https://extensions.gnome.org/extension/615/appindicator-support/) GNOME Shell extension from Canonical. Other applications don't have this issue. See this screenshot:

![Steam tray menu](https://user-images.githubusercontent.com/2339930/53288602-70438b00-378a-11e9-8892-9fdb2bbca201.png)

**Does this issue reproduce with native Steam**
Have not tested

nanonyme:
Has this worked before?

danrobinson78:
For what it's worth I'm seeing the same behavior under Plasma as well. Also on Fedora 29.

danrobinson78:
Per a comment on Reddit, I was able to symlink ~/.var/app/com.valvesoftware.Steam/.local/share/Steam to ~/.local/share/Steam and it caused the icon to appear.

[...]

PR #724, opened 2021-02-26
Update ~/.steam symlinks on launch

### LINKED ISSUE #480: "Browse local files" button in game properties in Steam client does not opens the file manager

gasinvein:
Read symlinks in `~/.steam` directory and, if XDG dirs prefix was changed, update symlinks to point to it.  
This enables migration for existing installations from `~` to `~/.var/app/com.valvesoftware.Steam` (and backwards, if needed) without the need to remove or update `~/.steam` directory manually.

- If `~/.steam` doesn't exist (i.e. a new installation), set the prefix to what was set in `FLATPAK_STEAM_XDG_DIRS_PREFIX` (`~/.var/app/com.valvesoftware.Steam` by default)
- If `~/.steam` exists 
  - If `FLATPAK_STEAM_UPDATE_SYMLINKS` set to true (false by default), update symlinks to what was set in `FLATPAK_STEAM_XDG_DIRS_PREFIX`
  - Otherwise, don't touch symlinks and set the prefix to the current one as determined by `~/.steam/root` symlink target

To test migration for existing installations:
`flatpak run --env=FLATPAK_STEAM_UPDATE_SYMLINKS=1 com.valvesoftware.Steam`

Fully fixes #480 and #322.

My primary concern here is current (previous) prefix detection. Any ideas how to detect it in a more reliable way are welcome.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -52,6 +52,7 @@ finish-args:
   - --env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES=
   - --env=STEAM_RUNTIME=
   - --env=FLATPAK_STEAM_XDG_DIRS_PREFIX=~/.var/app/com.valvesoftware.Steam
+  - --env=FLATPAK_STEAM_UPDATE_SYMLINKS=0
   - --env=SDL_VIDEODRIVER=
   - --env=DBUS_FATAL_WARNINGS=0
   - --env=SSL_CERT_DIR=/etc/ssl/certs
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -20,6 +20,19 @@ DEFAULT_CONFIG_DIR = ".config"
 DEFAULT_DATA_DIR = ".local/share"
 DEFAULT_CACHE_DIR = ".cache"
 FLATPAK_INFO = "/.flatpak-info"
+# List of symlinks under ~/.steam relative to xdg-data
+STEAM_SYMLINKS = {
+    "bin32": "ubuntu12_32",
+    "bin64": "ubuntu12_64",
+    "root": ".",
+    "sdk32": "linux32",
+    "sdk64": "linux64",
+    "steam": ".",
+}
+ALLOWED_XDG_DIRS_PREFIXES = [
+    os.path.expanduser("~"),
+    FLATPAK_STATE_DIR,
+]
 
 
 def read_flatpak_info(path):
@@ -38,6 +51,17 @@ def read_flatpak_info(path):
         "filesystems": flatpak_info.get("Context", "filesystems",
                                         fallback="").split(";")
     }
+
+
+def env_is_true(env_str: str):
+    if env_str.lower() in ["y", "yes", "true"]:
+        return True
+    if env_str.lower() in ["n", "no", "false"]:
+        return False
+    if env_str.isdigit():
+        return bool(int(env_str))
+    return None
+
 
 def read_file(path):
     try:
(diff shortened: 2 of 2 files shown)


nanonyme:
I would prefer adapting xdg variables instead of mutating data on disk. If there is a directory .steam, then its symlinks should be regarded as a point of truth.

gasinvein:
Existing installations will never transition, then.

nanonyme:
That sounds fine to me. The non-migrated state is not broken. I think forcing users to migrate is an unnecessary risk for little benefit.

gasinvein:
Ok, I'll add and a switch for allowing migration, disabled by default.

guihkx:
I did the (manual) migration as soon as #710 got merged. So far there are no (obvious) issues Steam itself and these games I play:

* CS:GO
* Bloons TD 6 (Proton)
* GTA:SA (Proton)
* Among Us (Proton)

The remaining question I have is: How do I revert this in case something breaks? Because I remember reading that this can cause issues with some Wine games (which ones?) running via Proton.

Thanks.

gasinvein:
>this can cause issues with some Wine games (which ones?) running via Proton

I don't know how this can possibly affect Proton games. `~` and `~/.var/app/com.valvesoftware.Steam` is effectively the same directory inside the sandbox, it even has the same inode. Can you point to a specific case when this breaks?

>How do I revert this in case something breaks?

Run Steam with `FLATPAK_STEAM_XDG_DIRS_PREFIX` set to `~` and `FLATPAK_STEAM_UPDATE_SYMLINKS` set to `1` to apply the migration backwards (doing this once should be enough).

guihkx:
>Can you point to a specific case when this breaks?

I was referring to [this comment](https://github.com/flathub/com.valvesoftware.Steam/issues/480#issuecomment-770204312). I'm also interested in knowing which Wine games might break (so I can test them).

>Run Steam with `FLATPAK_STEAM_XDG_DIRS_PREFIX` set to `~` and `FLATPAK_STEAM_UPDATE_SYMLINKS` set to `1` to apply the migration backwards (doing this once should be enough).

Thank you very much!

gasinvein:
> > Can you point to a specific case when this breaks?
> 
> I was referring to [this comment](https://github.com/flathub/com.valvesoftware.Steam/issues/480#issuecomment-770204312). I'm also interested in knowing which Wine games might break (so I can test them).

Hmm, I've read somewhere about Proton using a workaround for path length issue by mounting the steam library as windows drive `S:\`. Not sure if it's implemented (and/or enabled by default) already.

gasinvein:
Honestly, I don't like the fact that now we'll have two different kinds of installations: with the new xdg dirs prefix and with the old one.
But at least `XDG_*_HOME` env vars now will be consistent with `~/.steam` symlinks. That's better than before, I guess.

Issue #708, opened 2021-01-23
Psychonauts won't start


LoneFenris:
**Game information**
Psychonauts

**Distribution name and version where applicable**
Arch Linux

**Flatpak info**
Flatpak 1.10.0

default
host

**Problem description**
The game fails to properly start.
[psychonauts.log](https://github.com/flathub/com.valvesoftware.Steam/files/5860909/psychonauts.log)

**Does this issue reproduce with native Steam**
No

nanonyme:
Has it worked before or is this the first try?

LoneFenris:
I'm pretty sure I was able to run it a month or so ago.

nanonyme:
I would imagine `SDL_Init() failed: No available video device` is the fatal error here.

LoneFenris:
Probably. I wasn't sure if the looking for libsteam.so in my home directory errors were playing a part or not. Not sure why the `SDL_Init()` is failing regardless. 

For what it's worth, I'm running GNOME Wayland, though Steam seems to be using XWayland.

nanonyme:
Is this SDL 1.2 or SDL 2 game?

LoneFenris:
I'm not completely certain, but I think it's 1.2

gasinvein:
I think this lines may be relevant:
```
dlopen failed trying to load:
/home/lonefenris/.local/share/Steam/linux32/libsteam.so
with error:
/home/lonefenris/.local/share/Steam/linux32/libsteam.so: cannot open shared object file: No such file or directory
dlopen failed trying to load:
libsteam.so
with error:
libsteam.so: cannot open shared object file: No such file or directory
```
Looks like the game is trying to open `~/.local/share/Steam/linux32/libsteam.so` which isn't there (for me neither). Maybe it just broke after some Steam Client update?

LoneFenris:
@gasinvein That's a thought I had too. I did copy libsteam.so to the directory it was looking for it in, but there's no change in the error. It's my understanding that the flatpak isn't supposed to have access to that directory, anyway.

I did try it in the non-flatpak Steam now and it works. Not sure why the flatpak doesn't. Perhaps there's a library missing?

nanonyme:
@LoneFenris the changes are fast-tracked to stable. Can you give it a try now?

LoneFenris:
Yup. Runs now. Thanks.

PR #730, opened 2021-03-13
Fix environment variable masking with flatpak 1.10+

### LINKED ISSUE #729: Dungeons of Dredmor no longer launches
**Game information**
[Dungeons of Dredmor](https://store.steampowered.com/app/98800/Dungeons_of_Dredmor/)

**Distribution name and version**
Fedora 33

**Flatpak info**
Flatpak 1.10.1

default
host

**Problem description**
Dungeons of Dredmor used to work just fine under flatpak, but now won't launch.  The problem seems to be with SDL, as the line `Unable to init SDL: No available video device` shows up right after the game binary is launched.  I tried launching it with the env var `SDL_DEBUG` set to get more info, but that didn't give any further info.  [Here is the Steam output/log when attempting to run the game](https://github.com/flathub/com.valvesoftware.Steam/files/6131277/default.steam.log)

I've noted that the bourne shell wrapper script incorrectly identifies the platform as 32 bit, when I'm running a 64 bit machine, and but the same thing happens under native steam and the native steam version runs just fine, so it might be a problem with 32 bit SDL running under flatpak.  I thought forcing the wrapper script to use the 64 bit version would help, but it doesn't make things any better under flatpak and makes things worse under native Steam, so that's not it.

The game is using SDL 1.2, if that matters.

This might be a duplicate of issue #708, since it also has the same `SDL_Init() failed` error message.

Game works fine with native steam. When running natively it uses the versions of SDL shipped with the game, not the native SDL libraries in `/usr/lib`

nanonyme:
There's several environment variables which if leaked from host can break apps. Flatpak 1.10 has a behavioural change such that setting variables empty no longer removes them from environment but instead creates empty environment variables. We need to support both old and new flatpak so manifest needs to clear the variables in both old and new way.

Fixes https://github.com/flathub/com.valvesoftware.Steam/issues/729 https://github.com/flathub/com.valvesoftware.Steam/issues/708

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -40,6 +40,7 @@ finish-args:
   - --allow=bluetooth
   - --persist=.
   - --env=TZ=
+  - --unset-env=TZ
   - --env=LC_ADDRESS=C
   - --env=LC_COLLATE=C
   - --env=LC_MONETARY=C
@@ -48,12 +49,17 @@ finish-args:
   - --env=LC_NUMERIC=C
   - --env=LC_TELEPHONE=C
   - --env=ALSA_CONFIG_PATH=
+  - --unset-env=ALSA_CONFIG_PATH
   - --env=MESA_GLSL_CACHE_DIR=
+  - --unset-env=MESA_GLSL_CACHE_DIR
   - --env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES=
+  - --unset-env=STEAM_RUNTIME_PREFER_HOST_LIBRARIES
   - --env=STEAM_RUNTIME=
+  - --unset-env=STEAM_RUNTIME
   - --env=FLATPAK_STEAM_XDG_DIRS_PREFIX=~/.var/app/com.valvesoftware.Steam
   - --env=FLATPAK_STEAM_UPDATE_SYMLINKS=0
   - --env=SDL_VIDEODRIVER=
+  - --unset-env=SDL_VIDEODRIVER
   - --env=DBUS_FATAL_WARNINGS=0
   - --env=SSL_CERT_DIR=/etc/ssl/certs
   - --env=STEAM_EXTRA_COMPAT_TOOLS_PATHS=/app/share/steam/compatibilitytools.d


steam_wrapper/steam_wrapper.py
```diff
@@ -351,6 +352,11 @@ def configure_shared_library_guard():
     else:
         os.environ["LD_AUDIT"] = f"/app/links/$LIB/libshared-library-guard.so"
 
+def clear_empty_environment(current_info):
+    environment = current_info["environment"]
+    for key, value in os.environ.items():
+        if environment.get(key.lower()) == "":
```
gasinvein, line 358: Why convert to lowercase? I believe environment variables are case-sensitive and e.g. `TZ` and `tz` are actually different variables.

nanonyme: Python configparser quirk. Keys are lowercase.

gasinvein: Ah, right. I think we better fix that instead. Probably set `flatpak_info.optionxform = str` in `read_flatpak_info()` before loading the file would be enough?

LoneFenris:
As a note, I see that this only became an issue as of flatpak 1.10. Prior to that, env vars set to the empty string *would* unset them. The flatpak 1.10+ way of unsetting an env var seems to be using the --unset-env arg which writes to a new "unset-environment" entry in the metadata. 
From the documentation, it appears was done in order to allow packages to set empty environment variables.

nanonyme:
> As a note, I see that this only became an issue as of flatpak 1.10. Prior to that, env vars set to the empty string _would_ unset them. The flatpak 1.10+ way of unsetting an env var seems to be using the --unset-env arg which writes to a new "unset-environment" entry in the metadata.
> From the documentation, it appears was done in order to allow packages to set empty environment variables.

That should be fine. If the variables are not in your environment, then this algorithm will not do anything.

gasinvein:
This change will prevent users from adding their own overrides with env vars set to actual empty strings. But this seems a very unlikely use-case.

nanonyme:
> This change will prevent users from adding their own overrides with env vars set to actual empty strings. But this seems a very unlikely use-case.

Right. That's what I was thinking of. One option is changing from empty value to some specific value that signifies that the value should be removed from environment but I think it's a bit hacky.

steam_wrapper/steam_wrapper.py
```diff
@@ -351,6 +353,11 @@ def configure_shared_library_guard():
     else:
         os.environ["LD_AUDIT"] = f"/app/links/$LIB/libshared-library-guard.so"
 
+def clear_empty_environment(current_info):
+    environment = current_info["environment"]
+    for key, value in os.environ.items():
+        if environment.get(key) == "":
+            del os.environ[key]
```
gasinvein, line 360: Maybe spit a debug log here just in case?

nanonyme: Added logging, hopefully one that will not confuse the end user.

LoneFenris:
> > As a note, I see that this only became an issue as of flatpak 1.10. Prior to that, env vars set to the empty string _would_ unset them. The flatpak 1.10+ way of unsetting an env var seems to be using the --unset-env arg which writes to a new "unset-environment" entry in the metadata.
> > From the documentation, it appears was done in order to allow packages to set empty environment variables.
> 
> That should be fine. If the variables are not in your environment, then this algorithm will not do anything. Taking into a use a new option that won't work on older flatpak than 1.10 is not an option at this point. **I'm still trying to maintain compatibility up to 1.0.** We could have some improvement item to take new parameters into use in a couple of years.

Is that due to the version of flatpak you're building with? If not, and you're building with flatpak 1.10+, you could do both. That is - set the env var to an empty string with --env= *and* unset with --unset-env=. "unset-environment" is set up to take precedence obver env vars set in the Environment section in order to allow proper compatibility.

steam_wrapper/steam_wrapper.py
```diff
@@ -351,6 +353,12 @@ def configure_shared_library_guard():
     else:
         os.environ["LD_AUDIT"] = f"/app/links/$LIB/libshared-library-guard.so"
 
+def clear_empty_environment(current_info):
+    environment = current_info["environment"]
+    for key, value in os.environ.items():
+        if environment.get(key) == "":
+            logging.debug(f"Ensuring {key} not in environment")
```
gasinvein, line 360: I'd rather word it "Removing environment variable {key}". I mean, just say what we do, not why we do it.

nanonyme: I'm going to see if Flathub supports --unset-env. It allows doing this thing with no wrapper changes.

nanonyme:
> > > As a note, I see that this only became an issue as of flatpak 1.10. Prior to that, env vars set to the empty string _would_ unset them. The flatpak 1.10+ way of unsetting an env var seems to be using the --unset-env arg which writes to a new "unset-environment" entry in the metadata.
> > > From the documentation, it appears was done in order to allow packages to set empty environment variables.
> > 
> > 
> > That should be fine. If the variables are not in your environment, then this algorithm will not do anything. Taking into a use a new option that won't work on older flatpak than 1.10 is not an option at this point. **I'm still trying to maintain compatibility up to 1.0.** We could have some improvement item to take new parameters into use in a couple of years.
> 
> Is that due to the version of flatpak you're building with? If not, and you're building with flatpak 1.10+, you could do both. That is - set the env var to an empty string with --env= _and_ unset with --unset-env=. "unset-environment" is set up to take precedence over env vars set in the Environment section in order to allow proper compatibility. Older versions will ignore "unset-environment" and unset via the empty string. New versions obey "unset-environment".

Right, that worked perfectly. Reflected description and title.

LoneFenris:
Agreed about the cluttery-ness. That seems to be the way flatpak recommends doing things for now, though. :shrug:

PR #732, opened 2021-03-14
Fix metainfo updating

gasinvein:
flatpak-external-data-checker guesses that the last source in the manifest is the main one, which is wrong in our case, so we should explicitly mark the Steam source as main.


nanonyme (approved):
Sounds good. I never was a fan of the implicit behaviour

gasinvein:
I think the behavior changed after flathub/flatpak-external-data-checker#87. Before that, f-d-c-d would pick up any source with new version, which worked for Steam, but obviously was even more random.

PR #738, opened 2021-04-03
fixes on description metadata and adding Proton version required to play games

pablodav:
@nanonyme 

This is my proposal to update the description on metadata for steam flatpak version.

### DIFF
--- a/com.valvesoftware.Steam.metainfo.xml
+++ b/com.valvesoftware.Steam.metainfo.xml
@@ -12,19 +12,19 @@
     <p>This is a community package of Steam gaming plaform not officially supported by Valve. Report bugs through linked issue tracker.</p>
     <p>Steam is a popular platform for buying, downloading, and playing video games, and chatting with other players.</p>
     <p>Many games require an online purchase, but some popular games, such as Team Fortress 2, are free to play. When searching in the store, be sure to narrow results by the SteamOS/Linux operating system. Not all Linux games are compatible with your system, so make sure you check the requirements before purchasing games.</p>
-    <p>This version of Steam runs in flatpak, latest valve proton version to run windows games doesn't runs on flatpak, so you will need to install the complement of Proton Community Edition and change Steam Play preferences to use it</p>
+    <p>This version of Steam runs in flatpak. Starting from Valve Proton version 5.13 is not compatible with flatpak, so you will need to install the complement of Proton Community Build and change Steam Play preferences to use it</p>
   </description>
   <description xml:lang="pt_BR">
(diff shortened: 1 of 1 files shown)


nanonyme:
@gasinvein any views?

com.valvesoftware.Steam.metainfo.xml
```diff
@@ -9,12 +9,23 @@
   <summary xml:lang="pt_BR">Jogue jogos populares e os últimos lançamentos</summary>
   <summary xml:lang="es">Juega a los juegos populares y los últimos lanzamientos</summary>
   <description>
-    <p>This is a community package of Steam gaming plaform not officially supported by Valve. Report bugs through linked issue tracker.</p>  
+    <p>This is a community package of Steam gaming plaform not officially supported by Valve. Report bugs through linked issue tracker.</p>
     <p>Steam is a popular platform for buying, downloading, and playing video games, and chatting with other players.</p>
     <p>Many games require an online purchase, but some popular games, such as Team Fortress 2, are free to play. When searching in the store, be sure to narrow results by the SteamOS/Linux operating system. Not all Linux games are compatible with your system, so make sure you check the requirements before purchasing games.</p>
+    <p>This version of Steam runs in flatpak, latest valve proton version to run windows games doesn't runs on flatpak, so you will need to install the complement of Proton Community Edition and change Steam Play preferences to use it</p>
+  </description>
```
nanonyme, line 16: Capital first letter Valve and Windows at least.

gasinvein: It makes sense to clarify that versions starting from 5.13 can't work in flatpak yet.
Also explaining what Proton is is redundant here - people who don't know what it is yet, probably don't care if it works anyway.
Also I don't like the term "Community Edition". It implies that it's a customized fork like GE, which it isn't, instead it tries to stay as close to official Proton as possible.

pablodav: Thanks you for your comments,

Check how it looks now.

gasinvein: It's better now. Although I hope the official Proton won't be incompatible with flatpak forever, there is flatpak-side support waiting for review, so maybe add "yet" or "currently".

pablodav: done

gasinvein: Also please add a notice that the the community build is not supported by Valve, or reword it in a way so that it's clear that this build is provided by community, not Valve.

gasinvein: Or maybe it's not needed, since the description already says that this is flatpak is not supported by Valve.

nanonyme (approved):

PR #746, opened 2021-04-13
Make Proton extension autoinstalled

gasinvein:
Completely untested! But seems to work.

The idea is to create an additional extension point dedicated for Proton, with the same mount point, but matching its full ID without subdirectories. Its only purpose is to make flatpak autodownload the extension while keeping other compatibility tools optional.

`com.valvesoftware.Steam.CompatibilityTool.Proton` will match both extension points and be enabled twice, first time as normal and second time mounted over itself (I believe we don't care about order).

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -117,6 +117,12 @@ add-extensions:
     no-autodownload: true
     autodelete: false
 
+  com.valvesoftware.Steam.CompatibilityTool.Proton:
+    directory: share/steam/compatibilitytools.d/Proton
+    version: beta
+    versions: beta;test
+    autodelete: true
+
   com.valvesoftware.Steam.CompatibilityTool:
     subdirectories: true
     directory: share/steam/compatibilitytools.d
@@ -233,7 +239,7 @@ modules:
         install -Dm644 -t /app/etc ld.so.conf
         shared-library-guard-config-converter blocklist/* --outfile /app/etc/freedesktop-sdk.ld.so.blockedlist
         mkdir -p /app/lib{,32}/ffmpeg
-        mkdir -p /app/share/steam/compatibilitytools.d
+        mkdir -p /app/share/steam/compatibilitytools.d/Proton
         mkdir -p /app/utils /app/share/vulkan
         ln -srv /app/{utils/,}share/vulkan/explicit_layer.d
         ln -srv /app/{utils/,}share/vulkan/implicit_layer.d


com.valvesoftware.Steam.yml
```diff
@@ -117,6 +117,12 @@ add-extensions:
     no-autodownload: true
     autodelete: false
 
+  com.valvesoftware.Steam.CompatibilityTool.Proton:
+    directory: share/steam/compatibilitytools.d/Proton
+    version: beta
+    versions: beta;test
```
nanonyme, line 123: Is the `test` intentional?

gasinvein: Yes, it's same as for the multi-extension point.

gasinvein:
Also I'm a bit concerned about users who don't want Proton at all. Does flatpak at least provide a way to opt-out from Proton extension autoinstall?

nanonyme:
Kind of. There's `flatpak mask`

nanonyme:
Looks good to me otherwise.

HugLifeTiZ:
> Also I'm a bit concerned about users who don't want Proton at all. Does flatpak at least provide a way to opt-out from Proton extension autoinstall?

I think avoiding Proton is going to become increasingly difficult; games that are on the official Steam Play whitelist are going to show up in your library even if they run with Proton. I would say that auto-installing the extension would match how the distro package of Steam behaves, but there is a bit of a hitch; if something on the Steam Play whitelist requests Proton 5.13 specifically, it's not going to get one that works because the Flatpak Proton builds are currently not versioned, and the fact that they don't replace the regular Proton builds may also be a problem for whitelist games requesting 6.3 as well.

This won't matter much once [the work to allow subsandboxing](https://github.com/flatpak/flatpak/issues/3797) finishes up and lands in a stable release of Flatpak anyways, and I'm presuming the reason our Flatpak Proton builds don't have versioned names is because it's a temporary situation anyways.

nanonyme:
@HugLifeTiZ I would estimate that after there is a stable release of flatpak with subsandboxing, it will take 2-4 years until it has properly propagated through the ecosystem. We're here planning on something that will make things break less for users in the meantime.

nanonyme:
I don't like the idea of *replacing* official Proton builds. That goes IMHO already way too far with messing with Valve's solutions. Suggesting a different default is one thing (if that was possible) but I want the system to work so that if someone sets an override through Steam, then what they set is what they get.

HugLifeTiZ:
> @HugLifeTiZ I would estimate that after there is a stable release of flatpak with subsandboxing, it will take 2-4 years until it has properly propagated through the ecosystem. We're here planning on something that will make things break less for users in the meantime.

If we're expecting to need the Flatpak Proton builds to exist for two to four years, then we *really* ought to give them versioned names, or at least provide some other way to install multiple branches of Proton in parallel. Maybe we should just include all branches of Proton depending on Soldier in the Proton Flatpak, and the same for Proton-GE. As things are right now, the Flatpak version of Steam does not have a working Proton 5.13, and if you have games that need both 5.13 and 6.3, you're stuck.

> I don't like the idea of _replacing_ official Proton builds. That goes IMHO already way too far with messing with Valve's solutions. Suggesting a different default is one thing (if that was possible) but I want the system to work so that if someone sets an override through Steam, then what they set is what they get.

As it is right now, if someone sets an override for the game to run with official Proton 6.3 or the only Proton 5.13 present right now, what they get is a game that doesn't start anymore. We know the Proton versions based on Soldier just don't work at all, and won't work at all until subsandboxing is done, or Valve decides to patch Soldier so that it doesn't try to subsandbox if it's running in Flatpak. The current situation leaves Valve's official solutions entirely non-functional, and creates a duplicate list entry that may cause confusion. For that reason, I think that replacing official Proton builds is the most responsible thing to do, if feasible.

gasinvein:
> we _really_ ought to give them versioned names, or at least provide some other way to install multiple branches of Proton in parallel

That's not really a big deal. I've already [suggested](https://github.com/flathub/com.valvesoftware.Steam.CompatibilityTool.Proton/issues/4#issuecomment-812871352) to branch out the Proton community extension from that latest commit that had version 5.13, rename it (change flapak ID and paths), and submit to Flathub as a separate extension.
The actual issue here is that Steam won't pick up 3rd-party compatibility tool, regardless of its version, for a whitelisted game.

HugLifeTiZ:
> I've already suggested to branch out the Proton community extension from the latest commit that had version 5.13, rename it (change flapak ID and paths), and submit to Flathub as a separate extension.

Awesome, thanks!

> The only way around it is to mimic official Proton, something that we're not going to do.

Is it mainly just that it's difficult to do so considering each Proton branch is an official app with its own app ID and everything? I imagine that even with that, you could name the community build "Proton 6.3" and then what you end up with is two identical entries in your list of compatibility tools, which is *even more* confusing, and then whitelist games pick the wrong one anyways because they depend on the specific app. Or is this something that's actually possible, but we don't want to do so for a different reason?

gasinvein:
> Or is this something that's actually possible, but we don't want to do so for a different reason?

There are multiple reasons not to do this:

- Prefixes are not fully compatible, silently switching from official to community Proton (and backwards) will break some existing installations
- Doing so will interfere with using using official Proton, which should become usable in future flatpak releases
- As @nanonyme said, this goes way too far with messing with Valve's solutions

HugLifeTiZ:
> * Prefixes are not fully compatible, silently switching from official to community Proton (and backwards) will break some existing installations

This alone is a good enough reason IMO.

HugLifeTiZ:
> * Prefixes are not fully compatible, silently switching from official to community Proton (and backwards) will break some existing installations

This alone is a good enough reason IMO; I didn't even know this myself. I wonder which prefixes I have that broke from my switching back and forth between distro and flatpak Steam.

gasinvein:
Well, switching between may work, but I can't guarantee it won't break anything. For example, official Proton uses symlinks to reduce prefix size, while the community build doesn't (I think it's somehow compensated by ostree dedup) 'cause the symlinks wouldn't be resolvable outside the sandbox. Also fonts symlinks in prefixes point to different targets. And copying things over existing symlinks can go rogue at any time.

nanonyme:
Seems we're not merging so closing for clarity.

gasinvein:
Yeah, this is largely pointless to autoinstall the community build if it still has to be enabled manually. Should we investigate possibility to enable it automatically and reconsider this?

HugLifeTiZ:
I don't think it's necesssary; with Flatpak 1.11.x's sub-sandboxing, the standard builds of Proton that use Pressure Vessel work just fine now.

PR #750, opened 2021-04-27
Add per-app-dev-shm into manifest

nanonyme:
Part of https://github.com/flatpak/flatpak/issues/3797

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -38,6 +38,7 @@ finish-args:
   - --allow=multiarch
   - --allow=devel
   - --allow=bluetooth
+  - --allow=per-app-dev-shm
   - --persist=.
   - --env=TZ=
   - --unset-env=TZ


nanonyme:
This will keep failing until Flathub gets new enough flatpak to support the new switch.

smcv:
I was going to open this MR myself, thanks for getting there first!

The other possibility to make pressure-vessel work is `--device=shm`, but that weakens the sandbox (which is why I implemented `--allow=per-app-dev-shm`).

nanonyme:
@smcv I asked from @barthalion yesterday, they said what goes to PPA goes into Flathub servers in a couple of days. So if newer flatpak can be made available on build servers, then let's just rebuild this. If not, we need to think of something else.

nanonyme:
bot, build

nanonyme:
Tested. While flatpak 1.10 does show the new permission, it is silently ignored. Iirc minimum requirement for silently ignoring is something a lot older than 1.00 which is our minimum requirement anyway.

Issue #748, opened 2021-04-22
Insurgency segfaults on server browser


KhalilSantana:
Hello, sadly [Insurgency](https://store.steampowered.com/app/222880/Insurgency/) seems broken:

### SYMPTOMS
Game launches, and you can interact with any menu besides the ones containing the server browser UI (such as the "Multiplayer" or "Cooperative" tab).
Upon clicking any tab with the server browser UI, game freezes for a second and then quits with a SIGSEV.

### Attempted fixes
1. Validating game files. Always finds a single file which fails validation? Still segfaults
2. Deleting all GFX settings and workshop items Same segfault
2. Commenting out the LD_PRELOAD stuff on `steamapps/common/insurgency2/insurgency.sh`. The crash reporter doesn't crash, but the main game executable still crashes.
3. Using [the workaround listed on  Arch's wiki (launch flags)](https://wiki.archlinux.org/index.php/Steam/Game-specific_troubleshooting#Insurgency). Still segfaults.
4. Using the `steam`[package from Arch's multilib](https://archlinux.org/packages/multilib/x86_64/steam/) (which seems to default to steam-runtime). **This makes it work without any modification to launch options or to the game files.**

### Stack traces

**Main game:**
```
       Message: Process 160882 (insurgency_linu) of user 1000 dumped core.
                
                Stack trace of thread 440:
                #0  0x00000000f7dddb12 n/a (/app/lib/i386-linux-gnu/libc-2.31.so + 0x103b12)
                #1  0x00000000f7d5bc11 n/a (/app/lib/i386-linux-gnu/libc-2.31.so + 0x81c11)

```
**Crash reporter**
```
       Message: Process 157644 (crash_handler) of user 1000 dumped core.
                
                Stack trace of thread 351:
                #0  0x00000000f5c1a1b8 n/a (filesystem_stdio.so + 0x871b8)
                #1  0x00000000f5bd9934 n/a (filesystem_stdio.so + 0x46934)
                #2  0x000000000804c0b9 main (crash_handler + 0x40b9)
                #3  0x00000000f79acf7e n/a (/app/lib/i386-linux-gnu/libc-2.31.so + 0x1af7e)
```
### SYSTEM INFO
```
System:    Kernel: 5.11.15-arch1-2 x86_64 bits: 64 Desktop: KDE Plasma 5.21.4 Distro: Arch Linux 
[...]

PR #754, opened 2021-04-30
Use relative symlinks in /app/links/lib

RyuzakiKK:
When creating a new container with the new Flatpak sub-sandbox feature,
the original `/app` directory is binded as `/run/parent/app`.
But if we use absolute symlinks in `.../app/links/lib`, they will end up
being broken inside the new container because there `/app/lib` and
`/app/lib32` don't exist.


felipe19930:
bot, build com.valvesoftware.Steam

felipe19930:
I thought this could fix #748 by using the Steam Linux Runtime with flatpak 1.11.1, but instead the game crashes at the loading screen, i.e. even earlier than the default Flatpak runtime.

PR #764, opened 2021-05-23
Handle case where "~/.steam/root" does not exist

### LINKED ISSUE #763: Steam client don't start - No such file or directory ~/.steam/root
Steam client aren't starting. Before that it was crashed. And after reboot it wasn't started
I have reinstall Steam but it wasn't help for me

**System info**
```
OS: Arch Linux x86_64  
Kernel: 5.12.5-arch1-1  
DE: Plasma 5.21.5  
CPU: Intel i5-7400 (4) @ 3.500GHz  
GPU: AMD ATI Radeon RX 580
```

**Flatpak info**
```
flatpak --version  
Flatpak 1.11.1
```

```
flatpak --gl-drivers  
default  
host  
```

**Console log:**
```
/usr/bin/flatpak run --branch=stable --arch=x86_64 --command=/app/bin/steam-wrapper --file-forwarding com.valvesoftware.Steam @@u %U @@  
INFO:root:https://github.com/flathub/com.valvesoftware.Steam/wiki
  File "/app/bin/steam-wrapper", line 8, in <module>
    sys.exit(main())
  File "/app/lib/python3.8/site-packages/steam_wrapper.py", line 362, in main
    current_xdg_prefix = get_current_xdg_dir_prefix()
  File "/app/lib/python3.8/site-packages/steam_wrapper.py", line 298, in get_current_xdg_dir_prefix
    current_steam_root = os.readlink(os.path.expanduser("~/.steam/root"))
FileNotFoundError: [Errno 2] No such file or directory: '/home/vadim/.steam/root'
```

nanonyme:
Fixes https://github.com/flathub/com.valvesoftware.Steam/issues/763

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -295,7 +295,10 @@ def get_current_xdg_dir_prefix():
     if not os.path.isdir(os.path.expanduser("~/.steam")):
         logging.debug("~/.steam doesn't exists - cannot determine current prefix")
         return None
-    current_steam_root = os.readlink(os.path.expanduser("~/.steam/root"))
+    try:
+        current_steam_root = os.readlink(os.path.expanduser("~/.steam/root"))
+    except FileNotFoundError:
+        return None
     # FIXME we need a more reliable way to determine current prefix
     # here we assume that ~/.steam/root points to ~/.local/share/Steam
     # this will break if `steam` was first ran bypassing the wrapper


gasinvein:
Is this for a case when `~/.steam` dir exists, but `~/.steam/root` symlink doesn't?

nanonyme:
Unclear. I think this might also happen if the symlink is broken.

gasinvein:
Either case seems like broken installation to me. Should we try to handle it at all? If we should, maybe just check if `~/.steam/root` is a symlink and points to a directory (on line 295)?

nanonyme:
Afaik Steam will automatically repair these cases if we let it.

steam_wrapper/steam_wrapper.py
```diff
@@ -295,7 +295,10 @@ def get_current_xdg_dir_prefix():
     if not os.path.isdir(os.path.expanduser("~/.steam")):
```
gasinvein, line 295: Instead of catching readlink failure below, I'd rather do check the `~/.steam/root` symlink here. Something like this:
```suggestion
    steam_root_link = os.path.expanduser("~/.steam/root")
    if not (os.path.islink(steam_root_link) and os.path.isdir(steam_root_link)):
```

nanonyme: Makes sense.

PR #785, opened 2021-07-01
wip: modules: bunlde zenity instead of relying on the runtime

alatiera:
This is an experiemental patch

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1268

this commit message should be rewritten later


nanonyme:
Looks like the bundled functions as expected

nanonyme:
I'm closing this since there was debate on whether this change should be done at all in runtime side.

PR #809, opened 2021-09-22
Add `VulkanLayer` extensions bindirs to `PATH`

### LINKED ISSUE #811: Starting games with new mangohud runtime extension (org.freedesktop.Platform.VulkanLayer.MangoHud//21.08) fails
## Description
`org.freedesktop.Platform.VulkanLayer.MangoHud//21.08//stable` started showing the following message during `flatpak update`:
```
Info: (pinned) com.valvesoftware.Steam.Utility.MangoHud//stable is end-of-life, with reason:
   This extension was converted to a runtime extension, install org.freedesktop.Platform.VulkanLayer.MangoHud//21.08 instead
```
Removing `//stable` and installing `//21.08` does no longer allow games in Steam Flatpak to start. Simple `mangohud %command%` crashes every game on startup - while it was working before.

gasinvein:
Since MangoHud migrated to runtime extension, `mangohud` command is uawailable any more in Steam flatpak sandbox (see https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.MangoHud/issues/1#issuecomment-923994885).

To fix this, append each `.VulkanLayer` extension `bin` subdir to `PATH` via our wrapper.

Closes #811

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -33,6 +33,14 @@ ALLOWED_XDG_DIRS_PREFIXES = [
     os.path.expanduser("~"),
     FLATPAK_STATE_DIR,
 ]
+EXTENSIONS = {
+    "org.freedesktop.Platform.VulkanLayer": {
+        "mountpoint": "/usr/lib/extensions/vulkan",
+        "append-env": {
+            "PATH": [ "bin" ],
+        },
+    },
+}
 
 
 def read_flatpak_info(path):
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -346,6 +354,24 @@ def enable_discord_rpc():
...
+    installed_ext_ids |= set(flatpak_info["runtime-extensions"])
+    for ext_id, ext in EXTENSIONS.items():
+        for installed_ext_id in installed_ext_ids:
+            if not installed_ext_id.startswith(f"{ext_id}."):
+                continue
+            ext_basename = installed_ext_id[len(f"{ext_id}."):]
+            for env_var, subdirs in ext["append-env"].items():
+                for subdir in subdirs:
+                    dir_path = os.path.join(ext["mountpoint"], ext_basename, subdir)
+                    paths = [p for p in os.environ.get(env_var, "").split(os.pathsep) if p]
```
nanonyme, line 369: Why do we need to normalize it like this rather than just add at end the sequence?

nanonyme: Otherwise LGTM.

gasinvein: It's easier this way - we don't need to check is the env var is set or not.

gasinvein: Moved this to outer loop.

nanonyme (approved):

PR #824, opened 2021-10-31
Sync metainfo with upstream and add external game library notice

gasinvein:
Diverge less from the upstream metainfo.

Also adds a suggestion how to allow access to a game library on an external drive, as discussed on Matrix.


AsciiWolf:
LGTM! This may also fix [this](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1528) GNOME Software issue. :-)

gasinvein:
@AsciiWolf Looks like it didn't fix this?

nanonyme:
Maybe not but it looked correct change anyway

gasinvein:
Yes, the GS issue appears to be unrelated to metainfo, but to circular symlinks.

PR #832, opened 2021-11-14
Allow all extension branches (stable)

gasinvein:
This is #830 applied on top of git `master`.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -122,16 +122,16 @@ add-extensions:
   com.valvesoftware.Steam.CompatibilityTool:
     subdirectories: true
     directory: share/steam/compatibilitytools.d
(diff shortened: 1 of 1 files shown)


nanonyme:
Why not just PR from beta to master?

gasinvein:
Because it'll have merge conflict. And resolving it on Github will merge base (master) back to head (beta) branch, which is probably not what we want. That being said, if you want to merge `beta` into `master`, you'll need to do it locally, without going through PR.

nanonyme:
Wait, how did this situation happen to begin with? Will this PR fix merging beta to master?

gasinvein:
> Wait, how did this situation happen to begin with?

I believe with this change (it was intentionally only added to master but not to beta): #650

> Will this PR fix merging beta to master?

I think so.

PR #835, opened 2021-11-19
Fix downloading and mounting the VAAPI.Intel.i386 extension

doraskayo:
From the commit message:
```
This change aligns org.freedesktop.Platform.VAAPI.Intel.i386 to how
org.freedesktop.Platform.VAAPI.Intel is configured in the Platform
runtime.

It seems that the previous configuration meant that it was effectively
never mounted. This should enable 32-bit VAAPI to work with Intel GPUs.
```

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -98,12 +98,11 @@ add-extensions:
     directory: lib/i386-linux-gnu/dri/intel-vaapi-driver
     version: '21.08'
     versions: '21.08'
-    subdirectories: true
-    no-autodownload: true
     autodelete: false
+    no-autodownload: true
     add-ld-path: lib
-    download-if: active-gl-driver
-    enable-if: active-gl-driver
+    download-if: have-intel-gpu
+    autoprune-unless: have-intel-gpu
 
   org.freedesktop.Platform.ffmpeg-full:
     directory: lib/ffmpeg


doraskayo:
Tested with Steam Remote Play and it seems to work well.

However, it does require https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/6643 to be released before we can merge. The PR will remain in draft form until then.

com.valvesoftware.Steam.yml
```diff
@@ -227,8 +226,6 @@ modules:
         mkdir -p /app/bin
         mkdir -p /app/lib/i386-linux-gnu
         mkdir -p /app/lib/debug/lib/i386-linux-gnu
-        mkdir -p /app/lib/i386-linux-gnu/GL
-        mkdir -p /app/lib/i386-linux-gnu/dri/intel-vaapi-driver
```
gasinvein, line 231: These directories are useful in case when the `.GL32` extension is installed while `.Compat.i386` is not; it allows our wrapper and `steam` to run, while otherwise it hard fails early at flatpak level.

doraskayo: I see. I removed the commit.

doraskayo:
This is now ready because https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/6643 was released as part of `freedesktop-sdk-21.08.5`.

gasinvein (approved):
The extension point definition LGTM, yet I have no idea how to test if vdpau actually works.

PR #841, opened 2021-12-01
Check input device permissions

gasinvein:
Do an educated guess if proper udev rules are installed on host by checking ACL on `/dev/uinput`, and warn the user (just once) if they don't seem to be installed.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,3 @@
 [MESSAGES CONTROL]
 disable=logging-fstring-interpolation
+extension-pkg-allow-list=posix1e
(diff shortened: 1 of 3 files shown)


nanonyme (approved):
This looks fine to me. Have you tested this? Does it trigger on Fedora Silverblue?

gasinvein:
Sadly I don't have Silverblue, but tested it on regular Fedora Workstation, seems to trigger properly.

gasinvein:
Rewrote this using a class to handle user-facing messages.

Issue #813, opened 2021-10-04
Steam Friends is missing menu functionality due to missing lsof package


ppedro1:
**Game information**
It's not CSGO specific, but CSGO is a good example. 

**Distribution name and version where applicable**

Arch Linux 5.14.8-arch1-1

**Flatpak info**
flatpak --version: Flatpak 1.11.3
flatpak --gl-drivers: default, host

**Problem description**

I don't see the menu to join a friend's game or see friend's game info when right clicking on the friend on the friends list (or friend's chat). When running steam with `flatpak -vv run com.valvesoftware.Steam` I can see this output once I click a friend with right click:

```
sh: line 1: lsof: command not found
```

**Does this issue reproduce with native Steam**

This issue is resolved on the native/pacman installed steam by installing the `lsof` package.

PR #858, opened 2022-01-24
Add lsof to build. Resolves issue #813

mcd1992:
The steam friends menu uses lsof when checking if you can join a friend in their game. This adds lsof to the image.

`lsof -F u -i TCP@127.0.0.1:57132`


mcd1992:
Yes

PR #860, opened 2022-01-31
Check /dev/uinput exists before checking permissions

bbalp:
When devices=!all on flatpak overrides, /dev/uinput doesn't exist, which causes a crash when checking permissions.

```
  File "/app/bin/steam-wrapper", line 8, in <module>
    sys.exit(main())
  File "/app/lib/python3.9/site-packages/steam_wrapper.py", line 463, in main
    check_device_perms()
  File "/app/lib/python3.9/site-packages/steam_wrapper.py", line 128, in check_device_perms
    for entry in posix1e.ACL(file="/dev/uinput"):
FileNotFoundError: [Errno 2] No such file or directory: '/dev/uinput'
```

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -125,12 +125,14 @@ def read_file(path):
 def check_device_perms():
     has_perms = False
     logging.debug("Checking input devices permissions")
-    for entry in posix1e.ACL(file="/dev/uinput"):
-        if (entry.tag_type == posix1e.ACL_USER
-            and entry.qualifier == os.geteuid()
-            and entry.permset.write):
-            has_perms = True
-            break
+    uinput_path = Path("/dev/uinput")
+    if uinput_path.exists():
+        for entry in posix1e.ACL(file=uinput_path):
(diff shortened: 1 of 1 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -125,12 +125,14 @@ def read_file(path):
...
     has_perms = False
     logging.debug("Checking input devices permissions")
-    for entry in posix1e.ACL(file="/dev/uinput"):
-        if (entry.tag_type == posix1e.ACL_USER
-            and entry.qualifier == os.geteuid()
-            and entry.permset.write):
-            has_perms = True
-            break
+    uinput_path = Path("/dev/uinput")
+    if uinput_path.exists():
```
gasinvein, line 129: I'd rather `return None` if the path doesn't exist. It doesn't make sense to show the "no input device permissions" message if the device is unavailable.

gasinvein:
Thanks for the quickfix.
Since PRs should be merged to `beta` first (thus `beta` is default), I'll rebase your branch.

gasinvein:
bot, build

bbalp:
Thank you for your work and your quick response.

PR #865, opened 2022-02-05
Zypak subsandbox

doraskayo:
This PR so far only adds Zypak to the sandbox for easier experimentation.

### DIFF
--- a/modules.yml
+++ b/modules.yml
@@ -145,3 +145,13 @@ modules:
       - -Dshared_library_guard_config=/app/etc/freedesktop-sdk.ld.so.blockedlist
     sources:
       - "sources/shared-library-guard-git.json"
+
+  - name: zypak
+    sources:
+      - type: git
+        url: https://github.com/refi64/zypak
+        tag: v2021.09.1
+        commit: e09f37ca4ff09827434deba11f9b1f9e1c42dbaf
+        x-checker-data:
+          type: git
+          tag-pattern: ^v([\d.]+)$


nanonyme:
I think this is just the surface of the can of worms. We need to be able to launch pressure-vessel from inside zypak.

doraskayo:
> I think this is just the surface of the can of worms. We need to be able to launch pressure-vessel from inside zypak.

More work is certainly needed, and it's still not clear where exactly. This PR so far only adds Zypak to the sandbox for easier experimentation.

I'm mostly waiting for comments on https://github.com/ValveSoftware/steam-for-linux/issues/8373#issuecomment-1030699670, with @smcv likely being a key person in deciding the direction this is going to take.

smcv:
To avoid anyone having the wrong expectations: I do not have control over the proprietary parts of Steam, including steamwebhelper, and I do not have control over the timescale on which any change happens. I can make suggestions to Valve, but they will not necessarily implement them, and if they do, it will not necessarily be soon.

smcv:
> I think this is just the surface of the can of worms. We need to be able to launch pressure-vessel from inside zypak.

I am going to experiment with enabling Zypak for just the steamwebhelper, and not the rest of Steam. If I can get that to work, then I can ask Valve people whether they'd be willing to autodetect a Flatpak environment and wrap the steamwebhelper (only!) in Zypak in that case.

nanonyme:
@smcv is it possible for us as immediate mitigation to pass parameter to disable sandbox from wrapper script?

smcv:
> @smcv is it possible for us as immediate mitigation to pass parameter to disable sandbox from wrapper script?

Yes, if you run `steam -no-cef-sandbox` that will disable this.

smcv:
I'm also trying to get a change into the next Steam beta that will detect `/.flatpak-info` and automatically disable the sandboxing for now (equivalent to `-no-cef-sandbox`), which will put it back into the same situation that existed on the non-beta Steam branch.

smcv:
> Yes, if you run `steam -no-cef-sandbox` that will disable this.

#867

nanonyme:
Okay, let's have that be backup plan.

smcv:
The issue addressed by this PR has been split out from https://github.com/ValveSoftware/steam-for-linux/issues/8373 and is now tracked as https://github.com/ValveSoftware/steam-for-linux/issues/8421.

PR #875, opened 2022-03-02
Update required version

TheEvilSkeleton:
This MR updates the required minimum Flatpak version. `per-app-dev-shm` permission was recently announced on 1.11.1, per [this release](https://github.com/flatpak/flatpak/releases/tag/1.11.1).


gasinvein:
We still have a whole lot of users with Flatpak 1.10 and even earlier, mostly on Ubuntu and Debian ≤11. Bumping `require-version` to 1.11 is way to high IMO.

nanonyme:
Yeah, this is not okay at this point. Let's revisit in a year.

TheEvilSkeleton:
> We still have a whole lot of users with Flatpak 1.10 and even earlier, mostly on Ubuntu and Debian ≤11. Bumping `require-version` to 1.11 is way to high IMO.
> 
> > `per-app-dev-shm` permission was recently announced on 1.11.1, per [this release](https://github.com/flatpak/flatpak/releases/tag/1.11.1).
> 
> I believe unknown features are ignored by flatpak, so it's ok to have this finish-arg (not entirely sure, needs confirmation).

I just came across someone who had an issue with the Steam Flatpak where they couldn't even launch it. Every time they ran the command, it would output `bwrap: Can't mkdir parents for /run/host/os-release: Permission denied`. The only fix was to update it.

nanonyme:
Is there a bug report against flatpak about that?

gasinvein:
I'm pretty sure it's either a ZorinOS packaging or host setup issue. `per-app-dev-shm` doesn't have anything to do with `/run/host/os-release`, and the error looks like something that would affect all flatpaks, not just steam.

TheEvilSkeleton:
I see. I asked the person who had the issue and they told me that all the other apps were functional; only Steam wasn't.

nanonyme:
Wrong target branch.

TheEvilSkeleton:
> Started [test build 92456](https://flathub.org/builds/#/builders/32/builds/92456)

@nanonyme can you stop this build?

PR #879, opened 2022-03-06
Check if i386 extension are installed and warn user if not

### LINKED ISSUE #878: Error: You are missing the following 32-bit libraries, and Steam may not run: libc.so.6
Output of flatpak run com.valvesoftware.Steam:
`INFO:root:https://github.com/flathub/com.valvesoftware.Steam/wiki
INFO:root:Will set XDG dirs prefix to /home/krystian/.var/app/com.valvesoftware.Steam
DEBUG:root:Checking input devices permissions
WARNING:root:Missing permissions for input devices
INFO:root:Overriding TZ to Europe/Warsaw
steam.sh[2]: Running Steam on org.freedesktop.platform 21.08.11 64-bit
steam.sh[2]: STEAM_RUNTIME is enabled automatically
setup.sh[74]: Steam runtime environment up-to-date!
steam.sh[2]: Error: You are missing the following 32-bit libraries, and Steam may not run:
libc.so.6
steam.sh[2]: Can't find 'steam-runtime-check-requirements', continuing anyway
/home/krystian/.var/app/com.valvesoftware.Steam/.local/share/Steam/steam.sh: line 788: /home/krystian/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam: cannot execute binary file: Exec format error
`

using the beta branch with all the requirements installed, and the solution from https://github.com/flathub/com.valvesoftware.Steam/issues/805 still repeats.
I am running elementaryOS 6.1 with intel graphics.

The issue **does** reproduce with native steam.

gasinvein:
Closes #878

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -87,6 +87,18 @@ MSG_NO_INPUT_DEV_PERMS = Message(
     False,
 )
 
+MSG_NO_I386_COMPAT = Message(
+    "no-i386-compat",
+    "i386 compatibility extensions not installed",
+    (
+        "The following i386 compatibility flatpak extensions are not installed, "
+        "Steam may not work properly:<tt>\n{ext_ids}</tt>\n\n"
+        "Try running <tt>flatpak update</tt>; if it doesn't help, install the extensions manually:\n"
+        "<tt>flatpak install {to_install_refs}</tt>"
+    ),
+    True,
+)
+
 
 def read_flatpak_info(path):
     flatpak_info = configparser.ConfigParser()
@@ -94,6 +106,7 @@ def read_flatpak_info(path):
         flatpak_info.read_file(f)
     return {
         "flatpak-version": flatpak_info.get("Instance", "flatpak-version"),
+        "runtime": flatpak_info.get("Application", "runtime"),
         "runtime-path": flatpak_info.get("Instance", "runtime-path"),
         "app-extensions": dict((s.split("=")
                                 for s in flatpak_info.get("Instance", "app-extensions",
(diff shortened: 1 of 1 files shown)


nanonyme:
I think we should give GUI prompt and terminate app. This is a fatal error.

gasinvein:
> I think we should give GUI prompt and terminate app. This is a fatal error.

It won't be fatal when/if Steam Client itself becomes a 64-bit app.

nanonyme:
That is super unlikely. For now it will refuse to run at all with confusing errors.

gasinvein:
Well, if we're going to hard-abort execution, we should be super careful about false negative extension detection.

steam_wrapper/steam_wrapper.py
```diff
@@ -412,6 +429,19 @@ def enable_discord_rpc():
             os.symlink(src=src_rel, dst=dst)
 
 
+def check_extensions(flatpak_info):
+    _, _, _, runtime_branch = flatpak_info["runtime"].split("/")
+    installed_ext_ids = set(flatpak_info["runtime-extensions"]) | \
+                        set(flatpak_info["app-extensions"])
+    for ext_id in ["org.freedesktop.Platform.Compat.i386",
+                   "org.freedesktop.Platform.GL32.default"]:
+        if ext_id not in installed_ext_ids:
```
nanonyme, line 438: I would prefer these two be combined into same error.

gasinvein: The implementation is going to be rather ugly without a real template engine, but well, done.

nanonyme (approved):

steam_wrapper/steam_wrapper.py
```diff
@@ -55,13 +55,17 @@ class Message(t.NamedTuple):
     text: str
     always_show: bool
 
-    def show(self) -> bool:
-        logging.warning(self.title)
+    def show(self, subst: t.Dict[str, str] = None) -> bool:
```
nanonyme, line 58: **kwargs would also work here, I guess.

gasinvein: I thought about that, but using a single variable feels more future-proof. In case we'll need another arg, like `force_always_show` or something.

PR #896, opened 2022-04-28
chore(modules): update some modules to latest version

Lctrs:

### DIFF
--- a/modules-32bit.yml
+++ b/modules-32bit.yml
@@ -83,8 +83,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/gperftools/gperftools.git
-        commit: d8eb315fb18f6fb0d6efa923401f166343727bc6
-        tag: gperftools-2.8.1
+        commit: f7c6fb6c8e99d6b1b725e5994373bcd19ffdf8fd
+        tag: gperftools-2.9.1
 
   - name: shared-library-guard-32bit
     build-options:
--- a/modules.yml
+++ b/modules.yml
@@ -46,8 +46,8 @@ modules:
     sources:
       - type: git
         url: https://github.com/lsof-org/lsof
-        commit: 005e014e1abdadb2493d8b3ce87b37a2c0a2351d
-        tag: 4.94.0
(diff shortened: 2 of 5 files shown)


gasinvein:
Nitpick: updating modules seems like `chore` to me, not `feat`.

Lctrs:
Yeah I think you're right. But I don't think it's worth retriggering a build for this.

modules-32bit.yml
```diff
@@ -55,8 +55,6 @@ modules:
       arch:
         x86_64: *compat_i386_opts
     buildsystem: cmake-ninja
-    config-opts:
-      - -DGSTREAMER=ON
```
nanonyme, line 59: Why was this removed?

Lctrs: Because [GStreamer support has been removed in FAudio 22.02](https://github.com/FNA-XNA/FAudio/releases/tag/22.02).

Lctrs: (and therefore, the option doesn't exist anymore)

gasinvein: Ugh. I wonder why we need FAudio at all now, then.

Lctrs: I see now that I have missed the i386 counterpart.

nanonyme: Maybe we don't? It would be great to remove it, and the s-l-g overrides.

Lctrs: I'll remove the relevant commit from this PR and make a new one for it.

gasinvein: AFAIK it was needed exactly because it provided proprietary codecs support for Proton. If it doesn't provide it anymore, I'm not sure what else it's useful for.

Lctrs:
@nanonyme @gasinvein Here it is without the FAudio update

nanonyme:
fedc is fine if it works here.

modules-32bit.yml
nanonyme: it was removed through second PR so I'm resolving this.

Issue #901, opened 2022-05-13
Replace --socket=x11 permission with --socket=fallback-x11


rugk:
The GNOME Software considers applications requiring x11 as unsafe. There's a way to declare it uses Wayland, but can use X11 as a fallback, on desktop environments where Wayland doesn't run.

More info here:
https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/770#note_1136536

copied from https://github.com/flathub/org.gnome.Evolution/issues/49

nanonyme:
Blocked by https://github.com/ValveSoftware/steam-for-linux/issues/4924

doraskayo:
@gasinvein described the problem in doing this in https://github.com/flathub/com.valvesoftware.Steam/pull/902#issuecomment-1125862756, which is why it will likely never happen. This issue can probably be closed.

PR #902, opened 2022-05-13
Replace x11 by x11-fallback

rugk:
AFAIK it should work the same. It only falls back on non-Wayland systems on X11 then, which should be fine.

See/Fixes https://github.com/flathub/com.valvesoftware.Steam/issues/901


gasinvein:
`fallback-x11` causes x11 access to be disabled when wayland socket is available. This means that X11-only apps, including Steam Client itself and most games, won't work on Wayland session due to not being able to use Xwayland.

rugk:
Ah uff… that's bad… :eyes:

PR #876, opened 2022-03-02
Fix typos

TheEvilSkeleton:
This MR fixes some typos and is more formal.

FTR, the official term is "Flatpak application" and not "flatpak" or "flatpaks", per the [Flatpak docs](https://docs.flatpak.org/en/latest/introduction.html#:~:text=flatpak%20application%3A%20these%20are%20the%20applications%20the%20user%20installs%20via%20the%20flatpak%20command%20or%20via%20a%20different%20ui%20like%20gnome%20software%20or%20kde%20discover.).

PR #903, opened 2022-05-20
Revert "Fix typos"

nanonyme:
Reverts flathub/com.valvesoftware.Steam#876


gasinvein:
You probably could've just force push the `master` branch (instead of reverting), since it's not default and probably not protected.

nanonyme:
Iirc master branch *is* protected. It should be anyway if it's not since it's what the stable flatpak branch correlates with.

PR #905, opened 2022-05-20
Update required version

TheEvilSkeleton:
Continuation of https://github.com/flathub/com.valvesoftware.Steam/pull/875.


nanonyme:
Looks like based on flatpak stats we can proceed with this now.

PR #920, opened 2022-06-29
Switch com.valvesoftware.Steam beta to freedesktop-sdk 22.08beta

nanonyme:

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -1,6 +1,6 @@
 app-id: com.valvesoftware.Steam
 runtime: org.freedesktop.Platform
-runtime-version: '21.08'
+runtime-version: '22.08beta'
 sdk: org.freedesktop.Sdk
 command: steam-wrapper
 separate-locales: false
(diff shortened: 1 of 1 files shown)


com.valvesoftware.Steam.yml
```diff
@@ -107,16 +107,17 @@ add-extensions:
...
     autodelete: false
 
-  org.freedesktop.Platform.ffmpeg_full.i386:
-    directory: lib32/ffmpeg
-    add-ld-path: .
-    version: '21.08'
-    no-autodownload: true
-    autodelete: false
+#  TODO: freedesktop-sdk 22.08 doesn't currently have this
+#  org.freedesktop.Platform.ffmpeg_full.i386:
```
gasinvein, line 115: We don't need to remove extension point for a non-existent extension. It'll just stay empty, and should be installed and enabled when/if it becomes available.

nanonyme: Ah, cool.

nanonyme:
bot, build

nanonyme:
22.08beta glibc is broken. Trying to fix it through https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/8987. Let's revisit this later.

gasinvein:
bot, build

1player:
Trying out the new beta, upon startup it complains about:

* org.freedesktop.Platform.Compat.i386//22.08beta
* org.freedesktop.Platform.GL32.default//22.08beta

not being installed and fails with a missing libc error. Shouldn't these be installed automatically/as dependencies?

Installing them manually fixes it.

nanonyme:
This is common issue with beta app.

gasinvein:
> fails with a missing libc error

Ugh, this shouldn't be happening after #879 anymore. It should've shown a GUI error popup instead. I guess

1player:
It did show a GUI error, but then proceeded to load and broke.

nanonyme:
See https://github.com/flathub/com.valvesoftware.Steam/wiki/Using-beta-branch-of-this-app for installation instructions

PR #922, opened 2022-07-02
Remove lsb_release from app

nanonyme:
Looks like Steam can now read os-release directly so remove the
indirection


gasinvein:
Ugh, I wouldn't be so sure it's not needed anymore.

nanonyme:
It's not. I tested. In fact lsb_parser is buggy and results in corrupted output in sirveyu

nanonyme:
You can check from support page

PR #932, opened 2022-08-04
chore(modules): configure fedc

Lctrs:
Tested locally. Seems to work well.

### DIFF
--- a/sources/libusb-archive.json
+++ b/sources/libusb-archive.json
@@ -1,5 +1,11 @@
 {
     "type":"archive",
-    "url":"https://github.com/libusb/libusb/archive/v1.0.26.tar.gz",
-    "sha256": "a09bff99c74e03e582aa30759cada218ea8fa03580517e52d463c59c0b25e240"
+    "url":"https://github.com/libusb/libusb/releases/download/v1.0.26/libusb-1.0.26.tar.bz2",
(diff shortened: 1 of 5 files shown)


sources/libusb-archive.json
```diff
@@ -1,5 +1,11 @@
 {
     "type":"archive",
-    "url":"https://github.com/libusb/libusb/archive/v1.0.26.tar.gz",
-    "sha256": "a09bff99c74e03e582aa30759cada218ea8fa03580517e52d463c59c0b25e240"
+    "url":"https://github.com/libusb/libusb/releases/download/v1.0.26/libusb-1.0.26.tar.bz2",
```
Lctrs, line 3: According to a txt shipped with the [release](https://github.com/libusb/libusb/releases/tag/v1.0.26) on their github, we should use the .tar.bz2 file instead of the .tar.gz.

nanonyme: Sure, bzw is better compression format

nanonyme:
I don't understand the chore comment. Is this so ehow undesirable work?

Lctrs:
I just took the habit to follow the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/).

To me, fedc configuration is just an internal change and doesn't affect end users, hence chore classification. I can always drop that part if you want.

nanonyme:
This app is essentially all chore from that point of view. Just leave it out, it's unnecessary

Lctrs:
There you go.

PR #942, opened 2022-08-24
Update to freedesktop 22.08

orowith2os:
It seems to be available now, so probably best to update it.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -1,6 +1,6 @@
 app-id: com.valvesoftware.Steam
 runtime: org.freedesktop.Platform
-runtime-version: '21.08'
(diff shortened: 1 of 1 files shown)


orowith2os:
well that's interesting, because other apps are already using the 22.08 runtime from stable flathub.

nanonyme:
It's not yet released

orowith2os:
That's weird, because [there are other apps](https://github.com/flathub/dev.goats.xivlauncher/blob/master/dev.goats.xivlauncher.yml#L3) that are using it. I'll close this PR, then, since it doesn't seem to work.

nanonyme:
I don't know what's going on, that shouldn't work. The runtime is still being published as 22.08beta.

PR #951, opened 2022-09-04
Fix libdir for Worms W.M.D blacklist

heindsight:
My previous PR (#949) omitted the `/lib/` component of the path to `libstdc++.so.6`.


heindsight:
I've tested with the new build and this time it works correctly. Apologies for the earlier omission...

PR #970, opened 2022-09-23
Update content rating

### LINKED ISSUE #969: Age rating is inaccurate
The age rating, Adults Only, is wildly inaccurate as the application does not itself have any of the content listed. Instead, the application itself (and not the games available on it) should be rated. I took the liberty of generating a new OARS rating, which takes only the application into account, not any of its games, which I believe should be counted separately.

```xml
  <content_rating type="oars-1.1">
    <content_attribute id="social-chat">intense</content_attribute>
    <content_attribute id="social-info">intense</content_attribute>
    <content_attribute id="social-audio">intense</content_attribute>
    <content_attribute id="money-purchasing">intense</content_attribute>
  </content_rating>
```

salarua:
Fixes #969

### DIFF
--- a/com.valvesoftware.Steam.metainfo.xml
+++ b/com.valvesoftware.Steam.metainfo.xml
@@ -73,33 +73,10 @@
     <release version="1.0.0.62" date="2020-03-19"/>
     <release version="1.0.0.61" date="2019-04-18"/>
   </releases>
-    <content_rating type="oars-1.1">
-    <content_attribute id="violence-cartoon">intense</content_attribute>
-    <content_attribute id="violence-fantasy">intense</content_attribute>
-    <content_attribute id="violence-realistic">intense</content_attribute>
-    <content_attribute id="violence-bloodshed">intense</content_attribute>
-    <content_attribute id="violence-sexual">intense</content_attribute>
-    <content_attribute id="violence-desecration">intense</content_attribute>
-    <content_attribute id="violence-slavery">intense</content_attribute>
-    <content_attribute id="violence-worship">intense</content_attribute>
-    <content_attribute id="drugs-alcohol">moderate</content_attribute>
-    <content_attribute id="drugs-narcotics">moderate</content_attribute>
-    <content_attribute id="drugs-tobacco">moderate</content_attribute>
-    <content_attribute id="sex-nudity">intense</content_attribute>
-    <content_attribute id="sex-themes">intense</content_attribute>
-    <content_attribute id="sex-homosexuality">moderate</content_attribute>
-    <content_attribute id="sex-prostitution">moderate</content_attribute>
-    <content_attribute id="sex-adultery">moderate</content_attribute>
-    <content_attribute id="sex-appearance">intense</content_attribute>
-    <content_attribute id="language-profanity">intense</content_attribute>
-    <content_attribute id="language-humor">intense</content_attribute>
-    <content_attribute id="language-discrimination">intense</content_attribute>
+  <content_rating type="oars-1.1">
     <content_attribute id="social-chat">intense</content_attribute>
-    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-info">intense</content_attribute>
     <content_attribute id="social-audio">intense</content_attribute>
-    <content_attribute id="social-location">none</content_attribute>
-    <content_attribute id="social-contacts">none</content_attribute>
     <content_attribute id="money-purchasing">intense</content_attribute>
-    <content_attribute id="money-gambling">intense</content_attribute>
   </content_rating>
 </component>


gasinvein:
It's ongoing discussion whether age rating should apply only to the app itself, or to all the content accessible trough the app. I personally believe it should apply only to the app, especially when it has it's own age restrictions for content, but some people disagree.
My argument is as follows: of we apply age ratings to all content, then all the web browsers should be rated 21+ or whatever the absolute maximum is.

nanonyme:
I am deliberately ignoring this until someone tells me this change is fine. The original content rating was iirc pushed by Flathub staff.

barthalion (approved):
I don't know what the rationale was back when Steam has been added. These days, I ask submitters to apply OARS to the app, and not to the third party content that the app may offer. It's not like we warn users that Spotify may contain songs with vulgarisms, and so on.

salarua:
> My argument is as follows: if we apply age ratings to all content, then all the web browsers should be rated 21+ or whatever the absolute maximum is.

Exactly. Or GNOME Software itself being rated adults only.

nanonyme:
This is I think the original PR https://github.com/flathub/com.valvesoftware.Steam/pull/49

nanonyme:
I do in principle agree with the change.

gasinvein (approved):
Well, everyone seems to agree with this change.

nanonyme:
@barthalion @nedrichards was promoting Firefox should also have similar OARS to this. Possibly the primary reason it did not is Flathub decided not to package Firefox but instead waited for Mozilla to do it.

nanonyme (approved):

nanonyme:
And yes, the discussion also did cover whether GNOME Software should have very harsh OARS and the answer was not "heck no" but "it's a complicated matter".

gasinvein:
> whether GNOME Software should have very harsh OARS and the answer was not "heck no" but "it's a complicated matter".

Well, if we don't have a surely-correct answer, let's exclude surely-wrong ones. If we do apply age ratings to content, then GNOME Software itself should have the most harsh rating of all apps it has; this effectively makes access to _all_ apps gated by the most harsh rating. Seems surely-wrong to me.

com.valvesoftware.Steam.metainfo.xml
```diff
@@ -73,33 +73,10 @@
...
-    <content_attribute id="sex-appearance">intense</content_attribute>
-    <content_attribute id="language-profanity">intense</content_attribute>
-    <content_attribute id="language-humor">intense</content_attribute>
-    <content_attribute id="language-discrimination">intense</content_attribute>
+  <content_rating type="oars-1.1">
     <content_attribute id="social-chat">intense</content_attribute>
-    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-info">intense</content_attribute>
     <content_attribute id="social-audio">intense</content_attribute>
-    <content_attribute id="social-location">none</content_attribute>
```
gasinvein, line 100: Steam allows sharing location up to a country, if I'm not mistaken. So this attribute should be at least `mild`, I guess? Not sure if `mild` value is implemented for `social-location`.

com.valvesoftware.Steam.metainfo.xml
```diff
@@ -73,33 +73,10 @@
...
-    <content_attribute id="social-contacts">none</content_attribute>
```
gasinvein, line 101: IIRC Steam allows sharing email, correct me if I'm wrong.

com.valvesoftware.Steam.metainfo.xml
```diff
@@ -73,33 +73,10 @@
...
     <content_attribute id="money-purchasing">intense</content_attribute>
-    <content_attribute id="money-gambling">intense</content_attribute>
```
gasinvein, line 103: Can collectable cards or something else be considered gambling?

gasinvein (approved):
Ugh, I didn't manage to submit a review on the content attributes applied to Steam itself in time. But anyway, here it is, all comments are in question, so the current change still LGTM.

com.valvesoftware.Steam.metainfo.xml
nanonyme: Gambling: 
- Mild: Gambling on random events using tokens or credits
- Moderate: Gambling using fictional money
- Intense: Gambling using real money

com.valvesoftware.Steam.metainfo.xml
nanonyme: GeoIP based location is explicitly excluded.

com.valvesoftware.Steam.metainfo.xml
nanonyme: Right, you're correct, sharing email addresses counts as intense sharing of social-contacts.

nanonyme: Also possibly if Steam doesn't moderate communication
```
    <content_attribute id="social-chat">intense</content_attribute>
    <content_attribute id="social-audio">intense</content_attribute>
    <content_attribute id="social-contacts">intense</content_attribute>
```

com.valvesoftware.Steam.metainfo.xml
nanonyme: Steam explicitly bans money gambling on its platform. https://store.steampowered.com/eula/471710_eula_0
```
You must not implement any gambling, casino style games, or games of chance that involve any real currency, virtual currency, or Rec Tokens in any way.
```
So any game doing any of this is in violation of Steam policy.

nanonyme: As for games, it's tricky. There has been incidents where regulators have considered loot boxes et all as gambling.

com.valvesoftware.Steam.metainfo.xml
salarua: > Steam allows sharing location up to a country, if I'm not mistaken. So this attribute should be at least `mild`, I guess? Not sure if `mild` value is implemented for `social-location`.

Mild is not implemented, and Steam allows you to share your location up to the city level. Since it's not an exact address I elected to mark that as none, but that may be wrong.

com.valvesoftware.Steam.metainfo.xml
salarua: > Also possibly if Steam doesn't moderate communication
> 
> ```
>     <content_attribute id="social-chat">intense</content_attribute>
>     <content_attribute id="social-audio">intense</content_attribute>
>     <content_attribute id="social-contacts">intense</content_attribute>
> ```

Rating `social-chat` was tricky because chat does have automated filters which are turned on by default. But they don't catch everything and human mods usually aren't watching, so I marked it `intense` rather than `moderate`.

PR #980, opened 2022-10-06
add env to make (some) appimages run

### LINKED ISSUE #770: Steam flatpak is not able to launch Retroarch, which is available in steam as an Appimage
**Game information**
Retroarch - Steam Version

Linux Mint 19.3 Cinnamon

**Flatpak info**
 Flatpak version: Flatpak 1.11.1
 flatpak --gl-drivers: default host

**Problem description**
The Steam Flatpak version is no able to launch Retroarch, which is available in steam as an Appimage  

No, in native steam works as expected

soredake:
With this, appimages can now run, some will need new deps added to run (like https://github.com/flathub/net.lutris.Lutris/issues/299), some will need some arguments added to run (like `--no-sandbox` for electron appimages), but it's better that having appimages not working at all.

Fixes: https://github.com/flathub/com.valvesoftware.Steam/issues/770

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -22,6 +22,8 @@ finish-args:
   - --socket=pulseaudio
   - --share=network
   - --talk-name=org.gnome.SettingsDaemon.MediaKeys
+  # Needed for appimages
+  - --env=APPIMAGE_EXTRACT_AND_RUN=1
   # Steam in -steamos mode uses Network Manager for network settings, e.g.
   # connecting to wifi and so on.
   # In normal mode, it probably just gets info about current connection status.


gasinvein:
Do you have any examples of games that do and do not start working after this change?

soredake:
@gasinvein [i've tested some free games](https://github.com/LukeShortCloud/winesapOS/issues/328#issuecomment-1259281033), two required `--no-sandbox` flag to run (not related to appimages), supertux run of the box on wayland (i've tested in vm, so that may be the reason it crashed in x11), generally, non-electron games/apps should work. There are not too many appimages on steam, especially free, so can't test all of them.

gasinvein:
So, do I understand correctly that SuperTux is the only game that starts working OOTB?

soredake:
@gasinvein yes, but i think having at least some appimage's work OOTB better than none at all.

gasinvein:
Honestly, I'm not so sure if it worth it. Currently, all appimage-packed games don't work and it's easy to debug. But if mess with default appimage behaviour, most of these games still won't work but will crash for other, possibly different, reasons and require debugging each one individually.

soredake:
Another option is to merge it, and document how to fix non-working appimage games, which mostly will be electron-based which can be easy fixed with `--no-sandbox`.

soredake:
Another problem will be if user have 8Gb of ram or less, in this case tmpfs can be too small for big games to unpack into, I've encountered this with exoblast game appimage.

soredake:
Similar pr for lutris flatpak are merged https://github.com/flathub/net.lutris.Lutris/pull/300

HugKitten:
For those who want to get this to work, there is a solution. I'll use OVR as an example. First make sure you are not using the proton version, then find a run.sh file and edit it to something like this. Make sure to replace OVRAS.AppImage with the path and name of the app image:

```#!/usr/bin/env bash
unset LD_LIBRARY_PATH
./OVRAS.AppImage --appimage-extract-and-run

export APPIMAGE_EXTRACT_AND_RUN=1
./my.AppImage [...]

export APPIMAGE_EXTRACT_AND_RUN=1
env NO_CLEANUP=1 ./my.AppImage```

nanonyme:
Please don't post into closed pull requests.

PR #992, opened 2022-10-22
Simplify filesystem-manipulating code in wrapper 

gasinvein:
`shutil.copytree()` from the standard library has everything we need, namely the ability to use custom path exclude logic, so we don't really need to reinvent the wheel.


nanonyme:
Have you locally tested it? You can first bypass helper to create incorrect data to force migration

gasinvein:
> Have you locally tested it?

Yeah, superficially.

> You can first bypass helper to create incorrect data to force migration

That's what I did.

PR #1015, opened 2022-12-03
Add 22.08-extra to the GL32 extension point

doraskayo:
This allows the new Mesa extension with extra video codecs support to be mounted when installed.

CC: @nanonyme

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -2,7 +2,7 @@ app-id: com.valvesoftware.Steam
 runtime: org.freedesktop.Platform
 runtime-version: &runtime-version '22.08'
 x-gl-version: &gl-version '1.4'
-x-gl-versions: &gl-versions 22.08;1.4
+x-gl-versions: &gl-versions 22.08;22.08-extra;1.4
 sdk: org.freedesktop.Sdk
 command: steam-wrapper
 separate-locales: false


com.valvesoftware.Steam.yml
```diff
@@ -2,7 +2,7 @@ app-id: com.valvesoftware.Steam
 runtime: org.freedesktop.Platform
 runtime-version: &runtime-version '22.08'
 x-gl-version: &gl-version '1.4'
-x-gl-versions: &gl-versions 22.08;1.4
+x-gl-versions: &gl-versions 22.08;22.08-extra;1.4
```
gasinvein, line 5: `22.08-extra` should be preferred over `22.08` I suppose?
```suggestion
x-gl-versions: &gl-versions 22.08-extra;22.08;1.4
```

gasinvein: Nevermind, IIRC extension branch preference is right-to-left.

doraskayo: @gasinvein, the preference order is based on the extension's `priority` field as defined in its manifest. `22.08-extra` in this case has a higher priority than `22.08`. Confusingly, the lower the priority number; the higher the priority.

Even when two extensions share the same priority, confusingly yet again, the last of the two listed in the `versions` field would be the one preferred when mounting.

I suspect both cases are a result of the same ordering issue in Flatpak, but since they are age-old, they are likely here to stay.

doraskayo:
Seems to work well.

PR #1017, opened 2022-12-04
Read-only access to Discord IPC socket directory

Unrud:
Write access is not required to connect to the socket. It's only required to create it, which is done by Discord.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -35,7 +35,7 @@ finish-args:
   - --talk-name=org.freedesktop.Notifications
   - --filesystem=xdg-music:ro
   - --filesystem=xdg-pictures:ro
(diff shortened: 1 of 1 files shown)


com.valvesoftware.Steam.yml
```diff
@@ -35,7 +35,7 @@ finish-args:
   - --talk-name=org.freedesktop.Notifications
   - --filesystem=xdg-music:ro
   - --filesystem=xdg-pictures:ro
-  - --filesystem=xdg-run/app/com.discordapp.Discord:create
```
gasinvein, line 38: We need the `create` mode, because this way Discord can be started _after_ Steam. Otherwise, if the directory doesn't exist by the time Steam is started, it won't be created and Steam won't have access to it.

nanonyme:
Closing since as commented by @gasinvein, this breaks certain use cases.

Issue #556, opened 2020-04-28
Consider enabling --talk-name=org.freedesktop.Notifications by default


kon14:
I'm starting this thread as a discussion regarding adding --talk-name=org.freedesktop.Notifications to the finish-args and therefore enabling notifications by default.
This is related to #41. 

Enabling notifications could be considered a nuisance for some, but having it disabled by default only ends up confusing non-technical people attempting to casually enjoy their games, which I believe is one of the reasons why they choose to use Steam in the first place.

There are certain games that rely on this harmless functionality and having it disabled by default only makes certain people that are not particularly acquainted with Flatpak assume it's just broken and branding it as such.

Games like Dota2 rely on notify-send to inform players about their matches being ready.
Failure to join a match after queueing often results in people getting soft-banned from matchmaking, which is especially annoying for them and anyone queueing with them.

Since the game currently assumes notify-send is going to be provided by the host system without even checking for it the notification option in the settings is not even greyed out.
Therefore from a user's point of view they should be getting a notification for their matches, yet they end up with a matchmaking soft-ban instead, without ever being warned about needing to enable a permission in order to do so.

This is not exactly Flatpak's problem, but this barely makes a difference.
It should be noted that --no-talk-name has been added to Flatpak, therefore any user that wishes not to be bothered by notifications can easily override this permission without having to manually edit /var/lib/flatpak/overrides/com.valvesoftware.Steam anymore either.

PR #1062, opened 2023-03-27
Remove the use of org.freedesktop.Notification

A6GibKm:
Starting from libnotify 0.8, all calls go though the portal.


A6GibKm:
Needs testing.

nanonyme:
Needs testing with Dota 2 which was the original reason this was added.

nanonyme:
See https://github.com/flathub/com.valvesoftware.Steam/issues/556

A6GibKm:
Yeah this is probably not what we want, closing

PR #1071, opened 2023-04-18
wip: manifest: Tweak GST_PLUGIN_SYSTEM_PATH we set

alatiera:
The i386 compatibility extension gets mounted into /app/lib/i386-linux-gnu
which is then symlinked to from /usr/lib/i386-linux-gnu/ however
let's instead define the path directly to avoid having to go over
the symlink.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -67,7 +67,7 @@ finish-args:
   - --env=PROTON_DEBUG_DIR=/var/tmp
   - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
   - --env=XDG_DATA_DIRS=/app/share:/usr/lib/extensions/vulkan/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share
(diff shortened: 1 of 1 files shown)


com.valvesoftware.Steam.yml
```diff
@@ -67,7 +67,7 @@ finish-args:
   - --env=PROTON_DEBUG_DIR=/var/tmp
   - --env=XDG_CONFIG_DIRS=/etc/xdg:/usr/lib/x86_64-linux-gnu/GL:/usr/lib/i386-linux-gnu/GL
   - --env=XDG_DATA_DIRS=/app/share:/usr/lib/extensions/vulkan/share:/usr/share:/usr/share/runtime/share:/run/host/user-share:/run/host/share
-  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0:/usr/lib/i386-linux-gnu/gstreamer-1.0
+  - --env=GST_PLUGIN_SYSTEM_PATH=/app/lib/i386-linux-gnu/gstreamer-1.0:/app/lib/gstreamer-1.0:/app/lib32/gstreamer-1.0:/usr/lib/extensions/gstreamer-1.0:/usr/lib/x86_64-linux-gnu/gstreamer-1.0
```
gasinvein, line 70: I believe `/app/lib/gstreamer-1.0` should go before `/app/lib/i386-linux-gnu/gstreamer-1.0`. App bundle should take priority over the runtime; even though the Steam flatpak app doesn't actually include any gstreamer plugins currently, it serves as a copy-paste source for other multilib flatpak apps.

alatiera: Yes, this pr is mostly test something https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1565#note_1357689009

PR #1083, opened 2023-05-12
Add .GL32.Debug extension point

doraskayo:
Allows using the 32-bit Mesa debug extension.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -93,6 +93,15 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
+  org.freedesktop.Platform.GL32.Debug:
+    directory: lib/debug/lib/i386-linux-gnu/GL
+    version: *gl-version
+    versions: *gl-versions
+    subdirectories: true
+    no-autodownload: true
+    merge-dirs: *gl-merge-dirs
+    enable-if: active-gl-driver
+
   org.freedesktop.Platform.VAAPI.Intel.i386:
     directory: lib/i386-linux-gnu/dri/intel-vaapi-driver
     version: *runtime-version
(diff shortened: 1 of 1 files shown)


doraskayo:
Tested and verified that the extension and all the files are mounted in the right place. For example:
```
[📦 com.valvesoftware.Steam ~]$ ls -l {,/usr/lib/debug}/app/lib/i386-linux-gnu/GL/default/lib/dri/libgallium_dri.so*
-rwxr-xr-x 3 nfsnobody nfsnobody 32701600 Jan  1  1970 /app/lib/i386-linux-gnu/GL/default/lib/dri/libgallium_dri.so
-rw-r--r-- 3 nfsnobody nfsnobody 75519212 Jan  1  1970 /usr/lib/debug/app/lib/i386-linux-gnu/GL/default/lib/dri/libgallium_dri.so.debug
```

doraskayo:
@nanonyme, does it seem acceptable? The runtime now offers a .Debug extension for GL32.

com.valvesoftware.Steam.yml
```diff
@@ -92,6 +92,15 @@ add-extensions:
     download-if: active-gl-driver
     enable-if: active-gl-driver
 
+  org.freedesktop.Platform.GL32.Debug:
+    directory: lib/debug/lib/i386-linux-gnu/GL
+    version: *gl-version
+    versions: *gl-versions
+    subdirectories: true
+    no-autodownload: true
+    merge-dirs: vulkan/icd.d;glvnd/egl_vendor.d;OpenCL/vendors;lib/dri;lib/d3d;vulkan/explicit_layer.d;vulkan/implicit_layer.d
```
doraskayo, line 101: Should I make `merge-dirs` a variable to avoid repeating it?

doraskayo: Added.

PR #1096, opened 2023-06-09
Refactor file ignore logic to be re-usable by streamcmd_wrapper

### LINKED ISSUE #1028: Wrapper broken for steamcmd?
The `steamcmd` command doesn't seem to work. Version 1.0.0.75, Flatpak 1.14.1, Fedora 37

```sh
$ flatpak run --command=steamcmd com.valvesoftware.Steam
  File "/app/bin/steamcmd", line 8, in <module>
    sys.exit(main())
  File "/app/lib/python3.10/site-packages/steamcmd_wrapper.py", line 18, in main
    steam_wrapper.copytree(
AttributeError: module 'steam_wrapper' has no attribute 'copytree'
```

nanonyme:
This allows us to just use standard library copytree in both wrappers. Fixes https://github.com/flathub/com.valvesoftware.Steam/issues/1028

### DIFF
--- a/steam_wrapper/steam_wrapper.py
+++ b/steam_wrapper/steam_wrapper.py
@@ -193,6 +193,24 @@ def check_allowed_to_run(current_info):
     check_bad_filesystem_entries(current_info["filesystems"])
 
 
+class FileIgnorer:
+    def __init__(self, patterns: set):
+        self.patterns = patterns
+        assert not any(os.path.isabs(i) for i in self.patterns)
+
+    def __call__(self, root: str, names: t.List[str]) -> t.Set[str]:
+        ignored_names: t.Set[str] = set()
+        for name in names:
+            for ignored_pat in self.patterns:
+                if fnmatch.fnmatch(os.path.join(root, name),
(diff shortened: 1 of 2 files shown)


steam_wrapper/steam_wrapper.py
```diff
@@ -204,8 +222,7 @@ def __init__(self, source: str, target: str,
         assert os.path.isabs(self.target)
         self.ignore = ignore or set()
         self.rename = rename or []
-        self.no_copy = self.ignore | set(self.rename)
-        assert not any(os.path.isabs(i) for i in self.no_copy)
+        self.no_copy = FileIgnorer(self.ignore | set(self.rename))
```
gasinvein, line 225: Nitpick: a `self.ignorer` attribute name would be more sensible.

gasinvein (approved):
I can't really test this, but LGTM.

steam_wrapper/steam_wrapper.py
nanonyme: Let's just name it file_ignorer then. There is already an attribute called ignore and potential for typos is significant.

PR #1123, opened 2023-07-18
beta: Add gamescope VulkanLayer extension to PATH

orowith2os:
Necessary because of https://github.com/flathub/com.valvesoftware.Steam.Utility.gamescope/pull/150


gasinvein:
This shouldn't be needed. The wrapper handles adding extensions paths to `PATH` here:
https://github.com/flathub/com.valvesoftware.Steam/blob/5206bc6ea274faf94903442421764eaa656e8c3f/steam_wrapper/steam_wrapper.py#L435-L450

nanonyme:
@gasinvein let's confirm whether this is necessary or not. I want to do release from beta to stable soon and then make beta start providing 23.08beta.

gasinvein:
The `mangohud` command definitely did work last time I checked, I believe it should be no different for any other `.VulkanLayer` extension with binaries.

nanonyme:
Ok. I will close this then as unnecessary.

Issue #1078, opened 2023-05-04
Please consider making --command=steam go via the wrapper


smcv:
At the moment `/app/bin/steam` is Valve's `bin_steam.sh` (which users should never run directly in this Flatpak app), and `/app/bin/steam-wrapper` is the correct entry point for users (which does some necessary setup and then execs `/app/bin/steam`).

With https://github.com/flatpak/flatpak/pull/5404 and https://github.com/flatpak/flatpak/issues/5395 in mind, it might be better to have `/app/bin/steam` be the wrapper, and make it exec Valve's script as `/app/lib/steam/bin_steam.sh` (or via a symlink `/app/lib/steam/steam` after prepending `/app/lib/steam` to the `PATH`, if having it available in the `PATH` as `steam` is desirable).

That would make it less likely that users will get broken behaviour by running it with `--command=steam`. If https://github.com/flatpak/flatpak/pull/5404 gets merged, it would also allow the `steam` executable to be exported as a more explicit entry point `/var/lib/flatpak/com.valvesoftware.Steam+steam` alongside `/var/lib/flatpak/com.valvesoftware.Steam+steamcmd`.

`/app/bin/steamcmd` is already a symlink to `steamcmd-wrapper`, ensuring that `--command=steamcmd` goes via the wrapper as expected, instead of running Valve's code directly.

nanonyme:
@smcv this sounds fine, I actually considered it years back but wasn't sure if it was considered supported approach from Steam side.

nanonyme:
This is now changed in beta.

PR #1135, opened 2023-08-13
Make wrapper be /bin/steam

nanonyme:
As suggested by @smcv in #1078 let's make /bin/steam be wrapper script. This decreases probability of users accidentally bypassing wrapper and getting unexpected end results and facilicates some future use cases.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -5,7 +5,7 @@ x-gl-version: &gl-version '1.4'
 x-gl-versions: &gl-versions 23.08beta;23.08beta-extra;1.4
 x-gl-merge-dirs: &gl-merge-dirs vulkan/icd.d;glvnd/egl_vendor.d;OpenCL/vendors;lib/dri;lib/d3d;vulkan/explicit_layer.d;vulkan/implicit_layer.d
 sdk: org.freedesktop.Sdk
-command: steam-wrapper
+command: steam
 separate-locales: false
 rename-icon: steam
 rename-desktop-file: steam.desktop
@@ -92,7 +92,6 @@ add-extensions:
     merge-dirs: *gl-merge-dirs
     download-if: active-gl-driver
     enable-if: active-gl-driver
-    autoprune-unless: active-gl-driver
 
   org.freedesktop.Platform.GL32.Debug:
     directory: lib/debug/lib/i386-linux-gnu/GL
@@ -102,7 +101,6 @@ add-extensions:
     no-autodownload: true
     merge-dirs: *gl-merge-dirs
     enable-if: active-gl-driver
-    autoprune-unless: active-gl-driver
 
   org.freedesktop.Platform.VAAPI.Intel.i386:
     directory: lib/i386-linux-gnu/dri/intel-vaapi-driver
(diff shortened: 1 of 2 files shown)


com.valvesoftware.Steam.yml
```diff
@@ -180,13 +180,12 @@ modules:
         sed -i s:Exec=steam:Exec=/app/bin/steam-wrapper: steam.desktop
         sed -i s:/usr/bin/steam:/app/bin/steam-wrapper: steam.desktop
         desktop-file-edit --set-key=StartupWMClass --set-value=Steam steam.desktop
-        desktop-file-edit --remove-key=PrefersNonDefaultGPU steam.desktop
-        desktop-file-edit --remove-key=X-KDE-RunOnDiscreteGpu steam.desktop
```
doraskayo, line 184: This looks unrelated.

com.valvesoftware.Steam.yml
```diff
@@ -180,13 +180,12 @@ modules:
         sed -i s:Exec=steam:Exec=/app/bin/steam-wrapper: steam.desktop
         sed -i s:/usr/bin/steam:/app/bin/steam-wrapper: steam.desktop
```
doraskayo, line 181: Shouldn't this be something like this?
```suggestion
        sed -i s:/usr/bin/steam:/app/bin/steam: steam.desktop
```

com.valvesoftware.Steam.yml
```diff
@@ -180,13 +180,12 @@ modules:
...
         desktop-file-edit --set-key=StartupWMClass --set-value=Steam steam.desktop
-        desktop-file-edit --remove-key=PrefersNonDefaultGPU steam.desktop
-        desktop-file-edit --remove-key=X-KDE-RunOnDiscreteGpu steam.desktop
         sed -i s:/usr/lib/:/app/lib/: steam
         PREFIX=/app make install
         ln -sf /bin/true /app/bin/steamdeps
         cp /usr/bin/cmp /app/bin
         install -Dm644 -t /app/share/metainfo com.valvesoftware.Steam.metainfo.xml
+        ln -sf /app/bin/steam-wrapper /app/bin/steam
```
doraskayo, line 188: I feel that `/app/bin/steam` originally being a symlink to `/app/lib/steam/bin_steam.sh` is Steam's own implementation detail and can change at any time. For example, it could become a script or even a binary file in any update.

With that in mind, it's safer to move `/app/bin/steam` someplace else instead of assuming that it's a symlink that we can remove.

steam_wrapper/steam_wrapper.py
```diff
@@ -15,7 +15,7 @@
 
 
 FLATPAK_ID = os.getenv("FLATPAK_ID", "com.valvesoftware.Steam")
-STEAM_PATH = "/app/bin/steam"
+STEAM_PATH = "/app/lib/steam/bin_steam.sh"
```
doraskayo, line 18: If you follow my suggestion above, this should point at where we moved the original `/app/bin/steam`.

com.valvesoftware.Steam.yml
doraskayo: Also, `command: steam-wrapper` should be `command: steam`.

nanonyme: I suppose you're right. @smcv any idea why it uses full path to steam in Desktop Entry?

com.valvesoftware.Steam.yml
nanonyme: I don't think adding symlink buys anything. We could do test -x /app/lib/steam/bin_steam.sh to see the file exists in manifest so it fails in build time if this changes. Note that this path points to /app, it cannot arbitrarily change by without app rebuild.

nanonyme: Also, the path was suggested by @smcv, I am assuming they wouldn't have suggested it if it was likely to change.

com.valvesoftware.Steam.yml
```diff
@@ -177,16 +177,16 @@ modules:
...
+        sed -i s:/usr/bin/steam:/app/bin/steam: steam.desktop
         desktop-file-edit --remove-key=X-KDE-RunOnDiscreteGpu steam.desktop
+        desktop-file-edit --set-key=StartupWMClass --set-value=Steam steam.desktop
         sed -i s:/usr/lib/:/app/lib/: steam
         PREFIX=/app make install
         ln -sf /bin/true /app/bin/steamdeps
         cp /usr/bin/cmp /app/bin
         install -Dm644 -t /app/share/metainfo com.valvesoftware.Steam.metainfo.xml
+        ln -sf /app/bin/steam-wrapper /app/bin/steam
+        test -x
```
gasinvein, line 189: A leftover?

com.valvesoftware.Steam.yml
```diff
@@ -177,16 +177,16 @@ modules:
...
-        desktop-file-edit --remove-key=PrefersNonDefaultGPU steam.desktop
+        sed -i s:/usr/bin/steam:/app/bin/steam: steam.desktop
         desktop-file-edit --remove-key=X-KDE-RunOnDiscreteGpu steam.desktop
+        desktop-file-edit --set-key=StartupWMClass --set-value=Steam steam.desktop
         sed -i s:/usr/lib/:/app/lib/: steam
         PREFIX=/app make install
         ln -sf /bin/true /app/bin/steamdeps
         cp /usr/bin/cmp /app/bin
         install -Dm644 -t /app/share/metainfo com.valvesoftware.Steam.metainfo.xml
+        ln -sf /app/bin/steam-wrapper /app/bin/steam
```
gasinvein, line 188:
```suggestion
        ln -sf steam-wrapper /app/bin/steam
```

com.valvesoftware.Steam.yml
nanonyme: This is actually missing the actual path. Will fix.

PR #1171, opened 2023-09-29
Block `libtcmalloc_minimal.so.4` to avoid game crashes

doraskayo:
Source games started to crash in their vendored versions of `libtcmalloc_minimal.so.4` after the Freedesktop SDK runtime was updated to `23.08`.

Falling back to the version built [in the app manifest](https://github.com/flathub/com.valvesoftware.Steam/blob/291612cb6767069517412f65818e099b58a610ae/modules-32bit.yml#L65-L80) seems to avoid the crash.

Closes: https://github.com/flathub/com.valvesoftware.Steam/issues/1158

PR #1185, opened 2023-10-17
Fix for more source 1 games by blocking libtcmalloc.minimal.so.4

dreamyukii:
this pr will fix more source 1 games like Counter Strike Source and Black Mesa
based on https://github.com/flathub/com.valvesoftware.Steam/pull/1171

### DIFF
--- a/resources/blocklist/black_mesa.yml
+++ b/resources/blocklist/black_mesa.yml
@@ -1,0 +1,8 @@
+entries:
+  - basedir: "*(/*|/.*)/Black Mesa"
+    blocklists:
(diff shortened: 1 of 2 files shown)


sekalengrengginang:
it works, css and black mesa no longer crash

doogie544:
Thank you.  CS source *should* have been fixed with the last PR for this issue but when I was testing the games I somehow forgot to put css on the list so it was not added as it should have been.  I don't have Black Mesa so I was not able to test that one so was not aware of it being broken.

PR #1189, opened 2023-10-24
fix more source engine games that affected with malloc issue

dreamyukii:
this pr will fix Portal Reloaded, Half Life 2 Deathmatch,Half Life Source Deathmatch,and Entropy Zero 2 by blocking libtcmalloc_minimal.so.4


sekalengrengginang:
zombie panic source doesn't work, it throw "could not load library client"

Issue #1168, opened 2023-09-17
Steam doesn't start when launching it from app menu, but works with run command


Samuelinox:
When I try to start Steam from the app menu it gets indefinitely stuck in "loading user data" until killed from system monitor. The same doesn't happen though when i issue `flatpak run com.valvesoftware.Steam` from the terminal.

**Distribution name and version where applicable**
Debian GNU/Linux 12 (bookworm)

**Flatpak info**
Flatpak 1.14.4
gl drivers: default, host

These are the arguments used to launch steam from the app menu: 
`run --branch=stable --arch=x86_64 --command=/app/bin/steam --file-forwarding com.valvesoftware.Steam @@u %U @@`

I tried to add these to the terminal command:
`flatpak run --branch=stable --arch=x86_64 --command=/app/bin/steam --file-forwarding com.valvesoftware.Steam @@u %U @@`
and worked fine again.

Looking in the output of the issued command I do see this:

_Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 252
steam-runtime-launcher-service[252]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 252 exited
Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 374
steam-runtime-launcher-service[374]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 374 exited
Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 378
steam-runtime-launcher-service[378]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 378 exited
Steam Runtime Launch Service: steam-runtime-launcher-service keeps crashing on startup, disabling_

Could it be relevant?

user1-github:
Hey, I also have this "Unable to acquire bus name" error since I got the last Steam client update from September 11th, as I reported here: #1166.

I also use Debian 12 (with Gnome).

[...]

Issue #1166, opened 2023-09-15
Security errors in Gnome Logs since the last Steam client update


user1-github:
Since the last Steam client update from September 11, I've been getting this security error in Gnome Logs every time I launch Steam:

E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam".

It doesn't cause any noticeable issues, but is it something to worry about?

I'm on Debian 12.1.0 with Gnome 43.6,
Flatpak 1.14.4.

Zesko:
I confirm that.

```
steamwebhelper.sh[155]: Runtime for steamwebhelper: defaulting to /home/zesko/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_64/steam-runtime-heavy
steamwebhelper.sh[155]: Running under Flatpak, disabling sandbox
steamwebhelper.sh[155]: CEF sandbox already disabled
CAppInfoCacheReadFromDiskThread took 30 milliseconds to initialize
Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 317
steam-runtime-launcher-service[317]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 317 exited
Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 357
steam-runtime-launcher-service[357]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 357 exited
Steam Runtime Launch Service: starting steam-runtime-launcher-service
Steam Runtime Launch Service: steam-runtime-launcher-service is running pid 361
steam-runtime-launcher-service[361]: E: Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
Steam Runtime Launch Service: steam-runtime-launcher-service pid 361 exited
Steam Runtime Launch Service: steam-runtime-launcher-service keeps crashing on startup, disabling

```

Zesko:
here is the duplicate: https://github.com/flathub/com.valvesoftware.Steam/issues/1168

smcv:
> I've been getting this security error in Gnome Logs every time I launch Steam:
[...]

PR #1200, opened 2023-11-28
Add ownership to com.steampowered.PressureVessel.LaunchAlongsideSteam.*

burgurpun:
Add ownership to com.steampowered.PressureVessel.LaunchAlongsideSteam.* to resolve errors found in issues.

Unable to acquire bus name "com.steampowered.PressureVessel.LaunchAlongsideSteam"
https://github.com/flathub/com.valvesoftware.Steam/issues/1168#issuecomment-1741710694

I have not tested building.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -30,6 +30,7 @@ finish-args:
   - --talk-name=org.freedesktop.ScreenSaver
   - --talk-name=org.freedesktop.PowerManagement
   - --talk-name=org.freedesktop.Notifications
+  - --own-name=com.steampowered.PressureVessel.LaunchAlongsideSteam.*
   - --filesystem=xdg-music:ro
   - --filesystem=xdg-pictures:ro
   - --filesystem=xdg-run/app/com.discordapp.Discord:create


nanonyme:
I would like to have comment from @smcv whether this is actually an expected change.

nanonyme:
https://github.com/flathub/com.valvesoftware.Steam/issues/1166#issuecomment-1742799353

Looks like @smcv recommendation was adding `--own-name=com.steampowered.PressureVessel.*`

smcv:
`--own-name=com.steampowered.PressureVessel.*` is definitely proportionate, please add that as a minimum.

I think `--own-name=com.steampowered.*` would actually make sense. It is a domain name owned by Valve and used for the Steam app-store, and it will become necessary if users of the Flatpak app want to make use of a developer/debug interface provided by the container runtime, which makes per-game containers listen for commands from the host system on names like `com.steampowered.App123` (vaguely similar in principle to `flatpak enter`, but using IPC similar to `flatpak-spawn --host` rather than messing with namespaces).

barthalion:
Closing in favour of https://github.com/flathub/com.valvesoftware.Steam/pull/1208.

smcv:
> Closing in favour of #1208.

(It isn't closed yet, please could a maintainer or the author do that?)

barthalion:
Oops, must have pressed wrong button.

PR #1201, opened 2023-12-01
Fix parsing /.flatpak-info

TingPing:
The `/.flatpak-info` is a keyfile in the format described by the desktop-entry-spec.

It is *not* a format that the Python configparser module can correctly parse.

The most accessible and certainly correct implementation of this is GLib.Keyfile so I switched it to this.


TingPing:
It's just a dependency of pygobject. Installing it with pip didn't work.

PR #1206, opened 2023-12-07
use `id` instead of `app-id`

shubhisroking:
`app-id` is deprecated


nanonyme:
Please document into description how new flatpak-builder is needed for id.

PR #1346, opened 2024-09-18
Switch to 24.08

nanonyme:


bbhtt:
Looks like the error is from rapidjson which doesn't do releases, last release is from 2016. See https://github.com/Tencent/rapidjson/issues/2277 

Maybe backport https://github.com/Tencent/rapidjson/pull/719

Also let's remove libnotify as well, it's in 24.08

PR #1348, opened 2024-09-20
Propagate beta changes to stable

nanonyme:


nanonyme:
@bbhtt any clue what this conflict is about?

bbhtt:
It's in hwdata, I can't really edit these because the base is beta too. And the first few commits are already in master.

GitHub UI merge does not work well for these. Opened https://github.com/flathub/com.valvesoftware.Steam/pull/1349

nanonyme:
Well, it used to and I have fixed this multiple times already. The problem is if anything is directly merged to master without going through beta, entire workflow breaks permanently.

nanonyme:
The intent was to always do feature to beta to master to avoid beta and master diverging. Otherwise it's hard to say if all bug fixes are in all branches.

bbhtt:
It is diverged already, because I did rebases before. Although I can probably abuse my admin privileges to fix that by force pushing.

nanonyme:
Okay. Well, I'm no longer the maintainer so I don't really care. I was just explaining how the workflow was designed.

PR #1382, opened 2025-01-20
Include zlib-ng to improve performance

Eonfge:
Include zlib-ng with compatibility options. Ideally, this would make benchmarks with the Flathub version of Steam perform better.

---

For reference, distributions like Fedora already use zlib-ng so that their applications are more optimised on newer hardware. By also applying this change to this package, games might perform a little faster on hardware that supports advanced instructions.

Details about Fedora's migration:

- https://fedoraproject.org/wiki/Changes/ZlibNGTransition

Details about Zlib-ng:

- https://github.com/zlib-ng/zlib-ng

Similar discussion for the FreeDesktop SDK:

- https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1731

### DIFF
--- a/modules.yml
+++ b/modules.yml
@@ -3,6 +3,20 @@ buildsystem: simple
 build-commands: []
 modules:
 
+  - name: zlib-ng
+    buildsystem: cmake-ninja
+    config-opts:
+      - -DZLIB_ENABLE_TESTS=false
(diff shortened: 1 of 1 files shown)


Eonfge:
Got any feedback from the beta?

PR #1413, opened 2025-05-18
Access only to dri and input

gitrashad:
Since there is ```--device=input``` option includes game controllers  (see https://github.com/flatpak/flatpak/pull/5481), I think we have no reason to give all devices access to Steam. This PR only gives access ```dri``` and ```input``` to Steam.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -35,7 +35,8 @@ finish-args:
   - --filesystem=xdg-music:ro
   - --filesystem=xdg-pictures:ro
   - --filesystem=xdg-run/app/com.discordapp.Discord:create
-  - --device=all
+  - --device=dri
+  - --device=input
   - --allow=multiarch
   - --allow=devel
   - --allow=bluetooth


bbhtt:
input is available only to Flatpak >=1.16.0, Flatpak 1.14.x is still supported and will be for several years. So we can't add it or it will break the app for everyone on older Flatpak ie. majority of users on Debian/Ubuntu LTS.

It's written in the docs. Please do some research before doing these PRs.

gitrashad:
> input is available only to Flatpak >=1.16.0, Flatpak 1.14.x is still supported and will be for several years. So we can't add it or it will break the app for everyone on older Flatpak ie. majority of users on Debian/Ubuntu LTS.
> 
> It's written in the docs. Please do some research before doing these PRs.

It makes sense, thank you.

Erick555:
BTW: in PR [discussion](https://github.com/flatpak/flatpak/pull/5481), steam was mentioned as example app for which `--device=input` wouldn't be sufficient replacement for `--device=all` regardless of flatpak version:

> This is not going to work for hotplugged game controllers
...
> I know from my involvement in supporting Steam on Linux that users of game controllers typically consider non-working hotplug to be a showstopper issue
...
> That wouldn't be enough for Steam, but as you say, it would be a step away from --device=all for less demanding input device users.

PR #1514, opened 2026-02-17
Add MangoHud configuration to filesystem arguments

geeksville:
steam contains it's own mangohud defaults.  This change allows the user's preferences to be used.  So tools like [mango juice](https://flathub.org/en/apps/io.github.radiolamp.mangojuice) can work.


geeksville:
alas this change requires an [exception](https://docs.flathub.org/docs/for-app-authors/linter#exceptions).  let me know if you want it and I'll fire off the PR.

PR #1530, opened 2026-03-28
libspeechd, metadata and readme updates

user1-github:
- libspeechd: Update dotconf and libspeechd to the latest versions and move them to modules.yml. I think it's more logical to reserve the main manifest only to the Steam related modules.

- metadata: Remove instructions for accessing library on another drive. I see [this commit](https://github.com/flathub/com.valvesoftware.Steam/commit/940532b2ba3bb2a71bd4459e5533a1ab688e309e) has added filesystem permissions to secondary / external drives, so there is no need for these instructions anymore.

- Update minimum required Flatpak version in readme accordingly to the one set in `com.valvesoftware.Steam.yml`.


modules.yml
```diff
@@ -149,3 +149,52 @@ modules:
           url-template: https://github.com/pygobject/pycairo/archive/refs/tags/v$version.tar.gz
 
 # -- end pygobject --
+
+# -- libspeechd, used by Hypnospace Outlaw game --
+
+  - name: dotconf
```
bbhtt: why was this moved?

user1-github: I thought it's more logical to reserve `com.valvesoftware.Steam.yml` only to the Steam related modules.
But if you want them inside `com.valvesoftware.Steam.yml`, I can move them back.

bbhtt: yes it should be in the main manifest.

bbhtt: also just update the versions and add the x-checker para, the rest of the diff should be dropped.

user1-github:
bot, cancel

modules.yml
user1-github: done

PR #1021, opened 2022-12-13
Don't leak NetworkManager secrets into Steam app

nanonyme:
Documented to https://github.com/flathub/com.valvesoftware.Steam/wiki#App_starts_slower_than_native

PR #1563, opened 2026-08-10
Grant system-bus access to org.freedesktop.NetworkManager

psoetens:
Big Picture Mode queries NetworkManager over the system bus to display connectivity status. Without this permission, the D-Bus call is denied by the sandbox and BPM's network status (the 'wifi' icon in the title/status bar) is broken for every Flatpak Steam user.

### DIFF
--- a/com.valvesoftware.Steam.yml
+++ b/com.valvesoftware.Steam.yml
@@ -28,6 +28,8 @@ finish-args:
   - --system-talk-name=org.freedesktop.UDisks2
   - --talk-name=org.kde.StatusNotifierWatcher
   - --system-talk-name=org.freedesktop.UPower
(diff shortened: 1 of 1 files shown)


Erick555:
See https://github.com/flathub/com.valvesoftware.Steam/pull/1021 and https://github.com/flathub/com.valvesoftware.Steam/wiki#app-starts-slower-than-native

TLDR: you're free to set this locally if you want.

psoetens:
@Erick555 Thanks, I learnt something!  I guess we need an upstream fix to allow checks without exposing secrets. This should be a common use case I think. I understand if you want to close this for now...

Erick555:
There's upstream issue about this https://github.com/ValveSoftware/steam-for-linux/issues/7856

